PR #1, opened 2016-12-17
python 3 compatible print and pyc creation prevention

dryobates:
Tested on python 2.7.12 and 3.5.2


boxed:
Thanks!

boxed:
@dryobates Just a headsup that I've released a new version of mutmut with caching of mutations and different mutation IDs. I couldn't find an email to you so I have to use this place :P

PR #23, opened 2018-10-04
Handle annotated assignment in 3.6

worr:
In 3.6, annotated assignment was added, which takes the form of `a: int = 0`.
This would cause errors, since `:` is an operator, and its child would contain
nodes of type `annassign`.

We handle these the same way that we handle normal assignments, but by dropping
into the assignment logic at an offset in the expression we're handling.

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -169,8 +169,7 @@ def and_or_test_mutation(children, node, **_):
 
 
 def expression_mutation(children, **_):
-    assert children[1].type == 'operator'
-    if children[1].value == '=':
+    def handle_assignment(children):
         if getattr(children[2], 'value', '---') != 'None':
             x = ' None'
         else:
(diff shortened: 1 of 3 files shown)


boxed:
Great! Thanks for this. Do you need a release with this right now?

worr:
Nah, this can wait for your normal release cadence. Thanks for the offer though!! And thanks for the merge!

El 4 de octubre de 2018 11:45:23 p. m. PDT, "Anders Hovmöller" <notifications@github.com> escribió:
>Great! Thanks for this. Do you need a release with this right now?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/boxed/mutmut/pull/23#issuecomment-427262867

-- 
Enviado desde mi dispositivo Android con K-9 Mail. Por favor, disculpa mi brevedad.

PR #24, opened 2018-11-01
Make number_mutation more robust to different float formats

sonicxml:
1. the code would fail with floats using scientific notation, e.g. 1e-3 was what we had in our code, I believe. This is the reason for the try...except
2. Line 47 (the base = 10) was because our floats between 0 and 1, e.g. 0.1, were being set to base 8 by the if statement on line 37


boxed:
Thanks!

PR #48, opened 2018-11-26
Cache coverage data per filename

blueyed:
`coverage_data.lines()` was called a lot, but needs to be called only
once per file.

This adds caching for it.


boxed:
I did the rename after merge.

blueyed:
Cool.
Free free to cherry-pick / squash things in the future and amend them.

PR #50, opened 2018-12-12
WIP: Feature/refactoring Improved output

nklapste:
This is a stale branch containing potential improvements to the output of test execution (noting the mutation completed, the file that was mutated, and the mutant's status).

Output attempts to look similar to pytest output for familiarity.

General test output looks similar to this:

```
=========================== Starting Mutation Tests ===========================
Using test runner: python -m pytest -x
Running tests without mutations: cmd: python -m pytest -x
Obtained test execution time without mutations: 0.641284704208374
Collected 11 mutations from 1 file
slave\bar.py['e = 1'->'e = 2'] OK_KILLED
slave\bar.py['e = 1'->'e = None'] OK_KILLED
slave\bar.py['f = 3'->'f = 4'] OK_KILLED
slave\bar.py['f = 3'->'f = None'] OK_KILLED
slave\bar.py['d = dict(e=f)'->'d = dict(eXX=f)'] OK_KILLED
slave\bar.py['d = dict(e=f)'->'d = None'] OK_KILLED
slave\bar.py['g: int = 2'->'g: int = 3'] OK_KILLED
slave\foo.py['return a < b'->'return a <= b'] OK_KILLED
slave\foo.py['GLOBAL_THING = 123'->'GLOBAL_THING = 124'] OK_KILLED
slave\foo.py['GLOBAL_THING = 123'->'GLOBAL_THING = None'] OK_KILLED
slave\foo.py['if a > b:'->'if a >= b:'] BAD_SURVIVED
--- slave\foo.py
 +++ slave\foo.py
 @@ -12,7 +12,7 @@

  def bar(a, b):
      """some test junk function"""
 -    if a > b:
 +    if a >= b:
          return a
      else:
          return b

All mutation tests executed successfully!
================== KILLED:10 SUSPICIOUS:0 TIMEOUT:0 ALIVE:1 ===================
Executed 11/11 Mutations in 8.655 seconds
WARNING: Surviving mutants detected you should improve your tests
```

However the new changes to master seem to be in a difficult conflict with the current master and would require some effort to rebuild.

**Other Changes:**
* mutmut could now be natively executed within a windows environment
* much more docstring providing typing usage
* using `argparse` over `click` for simplicity (in this process config file usage was removed)
* abandonment of python2 support (this could be added back)
* Some misc files were also killed to simplify my development and build process they may be re-implemented as seen fit.

@boxed if you think any of these changes/refactors are useful let me know and I'll try to rebuild this for the new master.


nklapste:
Note: I don't expect this PR to be merged, but, rather idea's within it approved and re-implemented in a fresher branch.

boxed:
I don't think this kind of output is an improvement in the general case. Maybe having it behind a feature flag like "--ci" or "--verbose"? Because to me it will be just a mess if you're trying to actually use the tool. I used to have much more verbose output and cut it down over and over again because I was frustrated with it in actual use.

For example, printing killed mutants _at all_ was extremely frustrating when I tried running mutpy to compare. It just hides the actual information in a sea of irrelevant fluff. It's like it's optimized to look good on a background screen for a hollywood movie not actual use :P

So what I'm getting at is: what is your use case here? Because for interactive use to work through mutants this isn't improved output, it's radically worse in my opinion. I don't mean to sound too harsh, but I want to get a strong argument from you back :P

nklapste:
> 
> 
> I don't think this kind of output is an improvement in the general case. Maybe having it behind a feature flag like "--ci" or "--verbose"? Because to me it will be just a mess if you're trying to actually use the tool. I used to have much more verbose output and cut it down over and over again because I was frustrated with it in actual use.

This was mainly developed for a less interactive usage, such as running it within a CI program were logs are captured by 10,000 lines. The current design hides whats being executed and requires multiple commands to access that information later. 

With mutation testing I simply want an overview of what source files are currently tolerating fault injection, and with such information I can quick build up my tests to compensate for surviving mutants.

> 
> For example, printing killed mutants _at all_ was extremely frustrating when I tried running mutpy to compare. It just hides the actual information in a sea of irrelevant fluff. It's like it's optimized to look good on a background screen for a hollywood movie not actual use :P

I could see standard usage after an initial execution to simply use `findstr` (on windows) or `grep` to quickly filter output. 

e.g. given my original mutation test output noted in the original PR message we could filter to only see the 
mutants that survived with a simple piped command.
```
python ..\mutmut\mutmut slave | findstr BAD_SURVIVED
slave\foo.py['if a > b:'->'if a >= b:'] BAD_SURVIVED
```

I definitely agree that printing killed mutants can be quite a source of spam, maybe a filtering argument could be provided (e.g. `--print-mutants OK_KILLED, BAD_SURVIVED, ect...` . Or noting that existing CLI commands can be used for effective filtering, saving us from complicating mutmut.

Even then with this output format 200 killed mutants is only 200 lines of output. It also provides quick validation on what logic within a file is "fragile" (susceptible to fault injection).

> 
> So what I'm getting at is: what is your use case here? Because for interactive use to work through mutants this isn't improved output, it's radically worse in my opinion. I don't mean to sound too harsh, but I want to get a strong argument from you back :P

No harm done! Criticism is always good!

This was mainly spurred off from me using mutmut for testing graypy which includes many default args, globals, and requires a docker container to run full testing. Running this locally only provides a limited set of tests and thus, leads to many false positives.  Seeing the 🙁 number tick upwards without any indication what is being tested over is troubling. 

With an interactive CLI app going through each mutation and figuring out what has changed and validating that it is indeed not a weak mutation is tedious. The idea of me running 250 commands to see the state of each mutant that could be potentially a false positive scares me deeply.

**Some random ideas:**

Currently mutmut just has a running total displayed during execution. That could be retained at the bottom (as it is nice to know), but, I think having a log of what was mutated and its status at quick reach is useful.

Printing the full diff on a failure is somewhat noisy, would an output of just
`slave\bar.py[(@23)'f = 3'->'f = None'] OK_KILLED` be more concise and acceptable? Noting the line mutated, the actual mutation and its status?

To compare pitest (PIT) a Java mutation testing library overcomes output verbosity by giving detailed HTML reports that note quite effectively what was covered, what survived, and what was mutated.

Using emojis to note mutant status while fun is just gold plating. If you have to explain it is just another layer of complication.

boxed:
> This was mainly developed for a less interactive usage, such as running it within a CI program were logs are captured by 10,000 lines.

What does "by 10,000 lines" mean? In any case, yea I agree that the current output is pretty terrible for CI, that's why I suggested some basic CI-ish output. But crucially this should be being a feature flag because it would be terrible for the normal interactive use case imo.

> I definitely agree that printing killed mutants can be quite a source of spam, maybe a filtering argument could be provided (e.g. --print-mutants OK_KILLED, BAD_SURVIVED, ect... 

What would be the point in showing the killed mutants? Even behind a feature flag? It seems like it's just making mutmut more complex for a feature that is not useful.

> This was mainly spurred off from me using mutmut for testing graypy which includes many default args, globals, and requires a docker container to run full testing.

This does indeed seem pretty terrible! But just running it like that with docker and everything and just leaving it alone for a fews days should be ok no? How slow is it?

> but, I think having a log of what was mutated and its status at quick reach is useful.

You can run `mutmut results` while mutmut is still running actually. Maybe I should just write that in the output? I'm trying to make it pretty fool proof and intuitive. I don't want anyone to have to read docs if it's at all possible to avoid (compare with mutpy that reports mutants like "LOR" and a pseudo-diff that makes no sense :P ). 

I've been thinking of improving `mutmut show` to include a subcommand `mutmut show all` to display all diffs, maybe `mutmut show all <filename>` to show all surviving mutants for a specific file. 

> but, I think having a log of what was mutated and its status at quick reach is useful.

Hopefully you'll now think that, because of the above explanation about being able to access the detailed data during execution, this is irrelevant. 

> Using emojis to note mutant status while fun is just gold plating. If you have to explain it is just another layer of complication.

I do believe that the nomenclature of mutation testing is inherently confusing so using 🎉 for killed mutants I believe can actually help a lot.

nklapste:
> 
> 
> > This was mainly developed for a less interactive usage, such as running it within a CI program were logs are captured by 10,000 lines.
> 
> What does "by 10,000 lines" mean? In any case, yea I agree that the current output is pretty terrible for CI, that's why I suggested some basic CI-ish output. But crucially this should be being a feature flag because it would be terrible for the normal interactive use case imo.

Travis-CI normally has an output limit of 4MB. "10,000 lines" is based off of one of my past projects builds getting cut-off around 10,000 lines of captured stdout within Travis-CI. 
> 
> > I definitely agree that printing killed mutants can be quite a source of spam, maybe a filtering argument could be provided (e.g. --print-mutants OK_KILLED, BAD_SURVIVED, ect...
> 
> What would be the point in showing the killed mutants? Even behind a feature flag? It seems like it's just making mutmut more complex for a feature that is not useful.
> 
> > This was mainly spurred off from me using mutmut for testing graypy which includes many default args, globals, and requires a docker container to run full testing.
> 
> This does indeed seem pretty terrible! But just running it like that with docker and everything and just leaving it alone for a fews days should be ok no? How slow is it?

I usually let Travis-CI do all my docker based testing. Thus, spurring my requirement of mutmut to have more friendly output for non-interactive usage. 
> 
> > but, I think having a log of what was mutated and its status at quick reach is useful.
> 
> You can run `mutmut results` while mutmut is still running actually. Maybe I should just write that in the output? I'm trying to make it pretty fool proof and intuitive. I don't want anyone to have to read docs if it's at all possible to avoid (compare with mutpy that reports mutants like "LOR" and a pseudo-diff that makes no sense :P ).

I did not know of that feature. I'll check it out!

boxed:
Hmm.. well in the case of output limits you won't have much luck with this anyway. If
You aren't storing the mutmut cache you can't even use the output unless it's the full diffs as the IDs are now primary keys into the cache database. 

Don't know how that affects your use case. 

Personally I think mutmut in CI only makes sense conceptually in two modes: cache only (this exists) and differential (ie run it for changed lines in a PR only to verify that there are no new surviving mutants, there is an open issue for this).

nklapste:
Move major improvements into separate PRs (#54 and #53).

PR #53, opened 2018-12-17
Work with Windows 10

nklapste:
Allows `mutmut` to work within a Windows 10 environment.

### Other features:
* better details are now given when a subprocess times out
* Travis-ci windows environment tests!

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -387,10 +387,10 @@ def popen_streaming_output(cmd, callback, timeout=None):
         universal_newlines=True
     )
 
-    def kill(process):
+    def kill(process_):
         """Kill the specified process on Timer completion"""
         try:
-            process.kill()
+            process_.kill()
         except OSError:
             pass  # ignore
 
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -400,12 +400,17 @@ def popen_streaming_output(cmd, callback, timeout=None):
 
     while process.returncode is None:
         try:
-            output, errors = nonblock_read(process.stdout)
-            if output.endswith("\n"):
-                # -1 to remove the newline at the end
-                output = output[:-1]
-            line = output
-            callback(line)
+            if os.name == 'nt':
+                output = process.stdout.readline()
+            else:
+                output = nonblock_read(process.stdout)
+            if output:
+                if isinstance(output, bytes):
+                    output = output.decode("utf-8")
+                if output.endswith("\n"):
+                    # -1 to remove the newline at the end
+                    output = output[:-1]
+                callback(output)
         except OSError:
             # This seems to happen on some platforms, including TravisCI.
             # It seems like it's ok to just let this pass here, you just
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -18,6 +18,7 @@ from threading import Timer
 
 import click
 from glob2 import glob
+from nonblock import nonblock_read
 
 from mutmut import mutate_file, Context, list_mutations, __version__, \
     BAD_TIMEOUT, \
@@ -399,8 +400,11 @@ def popen_streaming_output(cmd, callback, timeout=None):
 
     while process.returncode is None:
         try:
-            # -1 to remove the newline at the end
-            line = process.stdout.readline()[:-1]
+            output, errors = nonblock_read(process.stdout)
+            if output.endswith("\n"):
+                # -1 to remove the newline at the end
+                output = output[:-1]
+            line = output
             callback(line)
         except OSError:
             # This seems to happen on some platforms, including TravisCI.
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -402,7 +402,7 @@ def popen_streaming_output(cmd, callback, timeout=None):
         try:
             process_.kill()
         except OSError:
-            pass  # ignore
+            pass
 
     # python 2-3 agnostic process timer
     timer = Timer(timeout, kill, [process])
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -430,7 +430,7 @@ def popen_streaming_output(cmd, callback, timeout=None):
             # won't get as nice feedback.
             pass
         if not timer.is_alive():
-            raise TimeoutError("subprocess timed out")
+            raise TimeoutError("subprocess running command '{}' timed out after {} seconds".format(cmd, timeout))
         process.poll()
 
     # we have returned from the subprocess cancel the timer if it is running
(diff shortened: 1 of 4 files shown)


boxed:
This breaks linux and mac though. The spinner in the released version updates every time you get a line of output. This change breaks that so you only get one spinner update for an entire test run of a mutant. I'm fine with falling back to the worse behavior when the advanced feature doesn't exist, but not by breaking the nicer feature when it's available.

nklapste:
> 
> This breaks linux and mac though. The spinner in the released version updates every time you get a line of output. This change breaks that so you only get one spinner update for an entire test run of a mutant. I'm fine with falling back to the worse behavior when the advanced feature doesn't exist, but not by breaking the nicer feature when it's available.

Went back to using `poll` and `readline` this should resolve this breaking issue.

boxed:
Nope, doesn't work. 

master.py:

```python
import subprocess
import shlex
from threading import Timer


def popen_streaming_output(cmd, callback, timeout=None):
    """Open a subprocess and stream its output without hard-blocking.
    :param cmd: the command to execute within the subprocess
    :type cmd: str
    :param callback: function to execute with the subprocess stdout output
    :param timeout: the timeout time for the processes' ``communication``
        call to complete
    :type timeout: float
    :raises TimeoutError: if the subprocesses' ``communication`` call times out
    :return: the return code of the executed subprocess
    :rtype: int
    """
    process = subprocess.Popen(
        shlex.split(cmd),
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        universal_newlines=True
    )

    def kill(process):
        """Kill the specified process on Timer completion"""
        try:
            process.kill()
        except OSError:
            pass  # ignore

    # python 2-3 agnostic process timer
    timer = Timer(timeout, kill, [process])
    timer.start()

    while process.returncode is None:
        try:
            # -1 to remove the newline at the end
            line = process.stdout.readline()[:-1]
            callback(line)
        except OSError:
            # This seems to happen on some platforms, including TravisCI.
            # It seems like it's ok to just let this pass here, you just
            # won't get as nice feedback.
            pass
        if not timer.is_alive():
            raise TimeoutError("subprocess timed out")

        process.poll()

    # we have returned from the subprocess cancel the timer if it is running
    timer.cancel()

    return process.returncode
    
def callback(line):
    print(line)
popen_streaming_output('python client.py', callback)
```

client.py: 

```python
import time

for _ in range(1000):
    print('foo')
    time.sleep(1)
```

This will not print the lines as they come but wait for the entire process to finish. At least on macOS. 

It was super hard to make the existing code work and there were tons of dead ends, you shouldn't try to fix it, but use another implementation if my version doesn't work.

nklapste:
> 
> 
> Nope, doesn't work.
> 
> master.py:
> 
> ```python
> import subprocess
> import shlex
> from threading import Timer
> 
> 
> def popen_streaming_output(cmd, callback, timeout=None):
>     """Open a subprocess and stream its output without hard-blocking.
>     :param cmd: the command to execute within the subprocess
>     :type cmd: str
>     :param callback: function to execute with the subprocess stdout output
>     :param timeout: the timeout time for the processes' ``communication``
>         call to complete
>     :type timeout: float
>     :raises TimeoutError: if the subprocesses' ``communication`` call times out
>     :return: the return code of the executed subprocess
>     :rtype: int
>     """
>     process = subprocess.Popen(
>         shlex.split(cmd),
>         stdout=subprocess.PIPE,
>         stderr=subprocess.PIPE,
>         universal_newlines=True
>     )
> 
>     def kill(process):
>         """Kill the specified process on Timer completion"""
>         try:
>             process.kill()
>         except OSError:
>             pass  # ignore
> 
>     # python 2-3 agnostic process timer
>     timer = Timer(timeout, kill, [process])
>     timer.start()
> 
>     while process.returncode is None:
>         try:
>             # -1 to remove the newline at the end
>             line = process.stdout.readline()[:-1]
>             callback(line)
>         except OSError:
>             # This seems to happen on some platforms, including TravisCI.
>             # It seems like it's ok to just let this pass here, you just
>             # won't get as nice feedback.
>             pass
>         if not timer.is_alive():
>             raise TimeoutError("subprocess timed out")
> 
>         process.poll()
> 
>     # we have returned from the subprocess cancel the timer if it is running
>     timer.cancel()
> 
>     return process.returncode
>     
> def callback(line):
>     print(line)
> popen_streaming_output('python client.py', callback)
> ```
> 
> client.py:
> 
> ```python
> import time
> 
> for _ in range(1000):
>     print('foo')
>     time.sleep(1)
> ```
> 
> This will not print the lines as they come but wait for the entire process to finish. At least on macOS.
> 
> It was super hard to make the existing code work and there were tons of dead ends, you shouldn't try to fix it, but use another implementation if my version doesn't work.

After further testing it seems this non-blocking readline method only works within a windows environ. I assume there must be some OS agnostic solution for a python readline call to be non-blocking. Worst case would be to retain your original code for Linux/osx and use this other method for windows. Either way I'll look into resolving this later this week.

boxed:
>  I assume there must be some OS agnostic solution for a python readline call to be non-blocking. 

I don't think there is. I did a lot of googling and trial and error to get to the code I have now. 

> Worst case would be to retain your original code for Linux/osx and use this other method for windows. Either way I'll look into resolving this later this week.

I think that is what we should do. 

I really appreciate your hard work on this! Keep at it!

mutmut/__main__.py
```diff
@@ -5,26 +5,29 @@
...
 from os.path import isdir, exists
 from shutil import move, copy
-from subprocess import Popen
-from threading import Thread
-from time import sleep
+from threading import Timer
 
 import click
 from glob2 import glob
+from nonblock import nonblock_read
```
boxed: That lib doesn't work on mac while the existing popen_streaming_output does. So no go.

nklapste: Thanks for the heads up, saves me some testing time!

nklapste:
Spinner should work for both windows and posix. For posix systems the method of spawning and reading from a subprocess is very similar to master.

Currently I'm having issues addressing some failures arising within the Travis build. 
* `pypy` build is encountering a requirements issue with `pytest-cov` (we could lock requirements but, I think this could be very limiting for a testing library)
* python3.6 tests are failing on travis, I cannot replicate these failures on my machine (Ubuntu 18.04.01 with python3.6.7)

boxed:
The coverage issue is solved now right? In any case this branch has conflicts now.

boxed:
We could also fib with the sqlite database to change the value. I think this makes more sense.

nklapste:
> 
> 
> We could also fib with the sqlite database to change the value. I think this makes more sense.

Take a look at #68 I feel that this feature might be needed in the future anyways as people may have systems have systems that could bog down under additional load. I've noticed on some of my lower spec machines the suspicious mutants can occur more often. Have the ability to increase this time limit can resolve this.

boxed:
Got some time to test this on macOS and it works now. This is pretty great!

PR #54, opened 2018-12-17
Feature/test all supported python versions

nklapste:
Rework of the `.travis.yml` `.setup.py` and tests to better operate over all versions of python and allow for easier CI testing via Travis-CI.

This removes the usage of `tox` within testing.

### Other Features:
* discovered and fixed python2.7 import issue of `zip_longest` (I believe this is already resolved as of now)
* requirements are now only defined within `setup.py`


nklapste:
> 
> 
>     1. I think you need to motivate this better. Using requirements.txt is standard practice and something I think is very good. Why are you removing it?

I'll revert this change.

> 
>     2. Why are you adding pylint in this PR? This seems unrelated.

The main goal of this PR was to improve testing of mutmut, using pylint seemed to be an easy win. 

The fact that it detected a undefined variable within the production code (see commit 5cce35d614d25455bc785ffeca15ffe067942c28) highlights its benefit.

> 
>     3. I think it doesn't make sense to test mutmut in 3.4, 3.5, 3.6 and 3.7. Just 3.7 and 2.7 should be fine.

Lots of people use 3.4, 3.5, and 3.6 and conformity between all three is not guaranteed. Testing to ensure run time on these versions should be mostly painless and provides a sanity check that everything is working.

It also ensures that our dependencies we install also conform to the older versions of python3.

> 
>     4. We should just remove pypy totally, that was basically just carried over from a previous code base, it's really not relevant. This is my bad.

If our tests say it "works" and were not using any python C libraries why drop support?

If we are going through the effort to support python2.7 we might as well support one of the reasons people still use it.

> 
>     5. please don't care about the silly 76 character column limit of PEP8. It just makes the test code uglier and harder to read.

That can be disabled in the `.pylintrc` or modified to have a larger limit.

nklapste:
Added back the usage of the requirements files within the `setup.py` though I had to remove the `-rrequirements .txt` within the the `test_requirements.txt` for the usage of `tests_require` to operate correctly within the `setup.py`.

boxed:
> The main goal of this PR was to improve testing of mutmut, using pylint seemed to be an easy win.

Sure, but it's a separate PR no?

> Lots of people use 3.4, 3.5, and 3.6 and conformity between all three is not guaranteed. Testing to ensure run time on these versions should be mostly painless and provides a sanity check that everything is working.

It's a risk-reward tradeoff. The risk that something works in 2.7 and 3.7 but not in 3.4/3.5/3.6 is extremely low. So low I don't think it's worth the CO2 produced by running the tests (or my time waiting for them).

> If our tests say it "works" and were not using any python C libraries why drop support?

I'm not saying drop support, I'm saying drop the testing because it's a waste of CPU cycles for something that will never actually break. It's the same logic as for 3.4 et al above.

nklapste:
> 
> 
> > The main goal of this PR was to improve testing of mutmut, using pylint seemed to be an easy win.
> 
> Sure, but it's a separate PR no?
> 
> > Lots of people use 3.4, 3.5, and 3.6 and conformity between all three is not guaranteed. Testing to ensure run time on these versions should be mostly painless and provides a sanity check that everything is working.
> 
> It's a risk-reward tradeoff. The risk that something works in 2.7 and 3.7 but not in 3.4/3.5/3.6 is extremely low. So low I don't think it's worth the CO2 produced by running the tests (or my time waiting for them).
> 
> > If our tests say it "works" and were not using any python C libraries why drop support?
> 
> I'm not saying drop support, I'm saying drop the testing because it's a waste of CPU cycles for something that will never actually break. It's the same logic as for 3.4 et al above.



> 
> 
> > The main goal of this PR was to improve testing of mutmut, using pylint seemed to be an easy win.
> 
> Sure, but it's a separate PR no?

I'll move pylint out into a separate PR.

> 
> > Lots of people use 3.4, 3.5, and 3.6 and conformity between all three is not guaranteed. Testing to ensure run time on these versions should be mostly painless and provides a sanity check that everything is working.
> 
> It's a risk-reward tradeoff. The risk that something works in 2.7 and 3.7 but not in 3.4/3.5/3.6 is extremely low. So low I don't think it's worth the CO2 produced by running the tests (or my time waiting for them).
> 
> > If our tests say it "works" and were not using any python C libraries why drop support?
> 
> I'm not saying drop support, I'm saying drop the testing because it's a waste of CPU cycles for something that will never actually break. It's the same logic as for 3.4 et al above.

I'll cut out testing pypy and older versions of python. Besides 2.7.

boxed:
Not sold on the pyenv clone from github. Seems like a security issue... but I see the advantage clearly! If you're done with this I'll merge.

nklapste:
Should be good to go for a re-review and merge.

* removed `pylint`
* removed the cloning of pypy for now
* resolved conflicts with new master
* removed testing for python3.4 and 3.5 (retaining 3.6 as that is used for the deployment build

.travis.yml
```diff
@@ -1,15 +1,44 @@
...
+script:
+  - python setup.py test
+after_success:
+  - codecov
+
+jobs:
+  include:
+    - python: 2.7
+#    - python: 3.4
+#    - python: 3.5
```
s-kovacevic: Maybe remove these rather than commenting out

mutmut/cache.py
```diff
@@ -1,13 +1,13 @@
-# coding=utf-8
+#!/usr/bin/env python
```
s-kovacevic: Do we really need shebang line in `cache.py`? I don't see `cache.py` being used as a standalone script. This stands for all shebangs in this PR, before adding them, ask yourself "is this file is ever going to be ran as a standalone".

mutmut/cache.py
```diff
@@ -21,9 +21,11 @@
     from itertools import izip_longest as zip_longest
 
 if sys.version_info < (3, 0):   # pragma: no cover (python 2 specific)
+    from itertools import izip_longest as zip_longest
```
s-kovacevic: Already fixed, see lines 18-21.

setup.py
```diff
@@ -1,8 +1,14 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+
+"""setup.py for mutmut"""
```
s-kovacevic: Generic comment, possibly auto generated, adds no value.

tests/test_main.py
```diff
@@ -1,6 +1,9 @@
-# coding=utf-8
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""pytests for :mod:`mutmut.__main__`"""
```
s-kovacevic: Again, no real value since the file is called `test_main`, stating the obvious.

tests/test_main.py
```diff
@@ -37,21 +46,17 @@ def test_foo():
 
 
 @pytest.fixture
-def filesystem():
-    shutil.rmtree('test_fs', ignore_errors=True)
-    os.mkdir('test_fs')
-    with open('test_fs/foo.py', 'w') as f:
-        f.write(file_to_mutate_contents)
+def filesystem(tmpdir):
+    foo = tmpdir.mkdir("test_fs").join("foo.py")
```
s-kovacevic: I like switching to tmpdir, step in the right direction.

tests/test_misc.py
```diff
@@ -1,3 +1,8 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""pytests for :mod:`mutmut.cache`"""
```
s-kovacevic: I'd rather rename the file to `test_cache` and remove this comment. Again shebang, encodings.

tests/test_mutation.py
```diff
@@ -1,3 +1,8 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""pytests for various mutations done by mutmut"""
```
s-kovacevic: Same as the above.

mutmut/cache.py
nklapste: We still have to do a python version check due to the usage of `text_type` should we move that import there (it explicitly hints to the usage of this import for backwards compatibility). Maybe not do this in this pr but maybe another.

s-kovacevic: Maybe for now just remove these lines so we don't duplicate the fix, but I see your point.

In my experience compat imports are done with:
```
try:
    ... 
except ImportError:
    ...
``` 
while
```
if sys.version_info > ...:
    IOWrapper = _BaseIOWrapper
```
for example is used to set attributes to the module.

https://github.com/python/cpython/blob/2e438cc2554495b28480a3ffe5cdf41b6ab823a0/Tools/gdb/libpython.py#L53-L56

https://github.com/python/cpython/blob/9fb051f032c36b9f6086b79086b4d6b7755a3d70/Tools/ssl/multissltests.py#L31-L34


 Maybe someone else can add some input on this one.

nklapste: Might as well not change it (thus retaining master's solution). I prefer less branching, but, catching `ImportError` is a powerful catch all for this issue.

s-kovacevic:
Sounds great to me!

Unit tests should run fast and should be available during the development, even though these tests are in the gray area (some of them are closer to integration tests) they are still fast enough to be used during the development and I feel that it is really important feature of unit tests. You want to run them on every save if possible, no need to wait for travis to tell you about a failed test after 10 minutes.
While responding to my review you mistakenly removed all of the encoding hints, not just the one in question so you made several commits trying to fix that, waited 4 minutes for each travis build to tell you there is one more file that needs to be fixed etc. My point is, if you had `make test` before the first push it would have been resolved in under two minutes with no commit history polution. :slightly_smiling_face: 

Anyway i'm rambling right now, if tox is back, this is good to be merged if you ask me.
Thank you for your patience, good job! :slightly_smiling_face:

nklapste:
* rebased (squashing to cleanup some commit spam)
* re-added the tox file for quick local testing 

Should be good to go for another test review!

Issue #37, opened 2018-11-13
OSError raised when waiting for pytest to finish


Etherealflux:
I checked out master/commit 63089e9 to try the new UI (pretty nice!) but ran into trouble:

mutmut appears to be waiting for a spawned process's `returncode` to be set, continually asking for a line from `stdout` until this condition holds. However, it appears that, at some point:

1. p.returncode is still None
2. stdout.readline() raises an OSError

Wrapping the readline() call in a `try/except pass` seems to fix the issue. It let the initial test run complete, and then started counting up the killed/survived totals.

Stacktrace:

```
⠙ Running...Traceback (most recent call last):
  File "/home/flux/anaconda3/bin/mutmut", line 5, in <module>
    main()
  File "/home/flux/.local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/flux/.local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/flux/.local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/flux/.local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/flux/anaconda3/lib/python3.6/site-packages/mutmut/__main__.py", line 57, in wrapper
    f(*args, **kwargs)
  File "/home/flux/anaconda3/lib/python3.6/site-packages/mutmut/__main__.py", line 236, in main
    baseline_time_elapsed = time_test_suite(swallow_output=not s, test_command=runner, using_testmon=using_testmon)
  File "/home/flux/anaconda3/lib/python3.6/site-packages/mutmut/__main__.py", line 431, in time_test_suite
    returncode = popen_streaming_output(test_command, feedback)
  File "/home/flux/anaconda3/lib/python3.6/site-packages/mutmut/__main__.py", line 306, in popen_streaming_output
    line = stdout.readline()[:-1]  # -1 to remove the newline at the end
OSError: [Errno 5] Input/output error
```

The bandaid I applied change line 306 to look like this:

```
        try:
            line = stdout.readline()[:-1]  # -1 to remove the newline at the end
        except OSError:
            pass
```

The `except` block is only hit once. `p.returncode` appears to be set properly afterwards, and the program carries on.

[...]

PR #57, opened 2018-12-25
Fix `popen_streaming_output` killing wrong process on Ubuntu

s-kovacevic:
Explanation is in the [issue thread](https://github.com/boxed/mutmut/issues/37#issuecomment-449871683).
Also removed unused imports because flake8 was yelling at me.

#37

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -5,6 +5,7 @@ from __future__ import print_function
 import itertools
 import os
 import sys
+import signal
 from datetime import datetime
 from difflib import unified_diff
 from functools import wraps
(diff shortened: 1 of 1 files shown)


mutmut/__main__.py
```diff
@@ -349,7 +350,7 @@ def timeout_killer():
             sleep(0.1)
             if (datetime.now() - start).total_seconds() > timeout:
                 foo['raise'] = True
```
s-kovacevic: Is there a reason why `global` is not used here. Like:
```python
should_raise = False
...
def timeout_killer():
     global should_raise
     should_raise = True
...
````

Issue #63, opened 2019-01-05
Running with --use-coverage fails


miketheman:
Hello! I am trying out mutmut today, having tried out cosmic-ray recently, and mutmut appears to be a simpler implementation, so thanks for making this!

My project uses pytest, and leverages coverage, and during my initial mutmut run, it found mutants in a function that I have a `pragma: no cover` set, so I wanted to use the flag to leverage only covered lines for mutation (there's also a couple of private methods that aren't exercised).

When invoking:

    mutmut run --use-coverage --paths-to-mutate unsilencer.py

I get this traceback:

```pycon
1. Using cached time for baseline tests, to run baseline again delete the cache file
Using coverage data from .coverage file
Traceback (most recent call last):
  File "/Users/miketheman/.local/share/virtualenvs/unsilencer-YmPJJ04A/bin/mutmut", line 5, in <module>
    main()
  File "/Users/miketheman/.local/share/virtualenvs/unsilencer-YmPJJ04A/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/miketheman/.local/share/virtualenvs/unsilencer-YmPJJ04A/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/miketheman/.local/share/virtualenvs/unsilencer-YmPJJ04A/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/miketheman/.local/share/virtualenvs/unsilencer-YmPJJ04A/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/miketheman/.local/share/virtualenvs/unsilencer-YmPJJ04A/lib/python3.7/site-packages/mutmut/__main__.py", line 59, in wrapper
    f(*args, **kwargs)
  File "/Users/miketheman/.local/share/virtualenvs/unsilencer-YmPJJ04A/lib/python3.7/site-packages/mutmut/__main__.py", line 285, in main
    coverage_data = read_coverage_data(use_coverage)
  File "/Users/miketheman/.local/share/virtualenvs/unsilencer-YmPJJ04A/lib/python3.7/site-packages/mutmut/__main__.py", line 484, in read_coverage_data
    coverage_data.read_file('.coverage')
AttributeError: 'CoverageSqliteData' object has no attribute 'read_file'
[...]

PR #64, opened 2019-01-07
Bugfix/coverage issue#63

nklapste:
Resolves the `AttributeError` being raised within Issue #63 due to deprecation of the `CoverageData.read_data` method.

Using the more modern file accessing api for the `coverage` library.


boxed:
This build failure is the same as the one you fixed before right? Can you just rebase and we can merge this?

PR #65, opened 2019-01-07
Bugfix/build test failures

nklapste:
Locking the test requirements of `pytest` and `pytest-cov` to work with the current tests.

* newer versions of `pytest` (4.0.0+) fail with the current testing suite of `mutmut`.
* `pytest-cov-2.6.1` creates a requirements issue with the testing of `pypy`

This is mainly a hotfix to keep our build stable untill we can properly investigate and address the issues with the newest version of `pytest`.


nklapste:
On further research this seems to be isolated only to the Travis environment. Using both pytest 4.1.0 and pytest-cov 2.6.1 within a local Linux machine still passes tests.

boxed:
Locking versions is a good idea anyway. Let's do this.

PR #68, opened 2019-01-09
Feature/modify base line test times

nklapste:
Adds the CLI options `--test-time-base` and `--test-time-multiplier` to allow users to modify the time thresholds before a mutant test run is considered `Suspicious 🤔`.

Due to these options, tests validating the detection of `Suspicious 🤔` mutants can be made easily (and were made).


mutmut/__main__.py
```diff
@@ -187,12 +195,18 @@ def climain(command, argument, paths_to_mutate, backup, runner, tests_dir,
     show [mutation id]\n
         Show a mutation diff.\n
     """
-    sys.exit(main(command, argument, paths_to_mutate, backup, runner, tests_dir,
-            swallow_output, use_coverage, dict_synonyms, cache_only, version,
-            suspicious_policy, untested_policy))
+    if test_time_base is None:  # click sets the default=0.0 to None
```
nklapste: If you provide `0` or `0.0` as an argument to either `--test-time-multiplier` or `--test-time-base` `click` will interpret this as `None`. This resolves this odd behavior.

PR #71, opened 2019-01-10
WIP: Feature/pylint tests

nklapste:
Adding a simple setup.py command `python setup.py lint` to execute `pylint` over the code. 

Also adds a test stage `lint` to the Travis build.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,407 @@
+[MASTER]
+
+# Specify a configuration file.
+#rcfile=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+init-hook='sys.path.append(os.path.abspath(os.path.curdir))'
+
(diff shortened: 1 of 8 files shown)


boxed:
Why not just execute the existing tox command in travis?

nklapste:
> 
> 
> Why not just execute the existing tox command in travis?

That is true. I always liked how `pylint` offered some cool convention checks, maybe we could run both as a heavy linting run? Though, with all our py2-3 patches this may be cumbersome.

I find running tox somewhat overkill on Travis as your making another venv (which in travis is somewhat redundant, thus, slowing down your builds). However, I do love using tox locally. 

Alternatively, we still use flake8, and later I'll code up a PR addressing some issues a local pylint noted.

nklapste:
Adopted Flake8 for linting instead.

Issue #61, opened 2019-01-04
Support for HTML reports


s-kovacevic:
While looking at mutation libraries for other languages I found out that PIT (Java) has a really nice HTML report output. I think that feature like this would make mutmut easier to grasp especially for people new to mutation testing.
Also mutmut is difficult to work with when you have a lot of mutations, like:
`mutmut show 58`
`mutmut show 59`

It seems like a lot of work, but I want to hear opinions do you guys think this would be worth doing. 
Any other ideas that would improve result reading experience are welcome.

![image](https://user-images.githubusercontent.com/10282513/50667682-b403ea80-0fba-11e9-9759-e2532c256e09.png)

![image](https://user-images.githubusercontent.com/10282513/50667514-e6611800-0fb9-11e9-9bf1-4cdbd8697cd7.png)

boxed:
Hmm, that doesn't look _that_ nice to me :P I think we can make something nicer very simply. Should be just looping over the cache.

The big downside to me is that I won't be able to brag that mutmut is ~1000 lines of code anymore (it's already a bit over that). This is not a very significant downside :P

nklapste:
Having reports similar to PIT would definitely meet some of my use cases with larger projects!

Maybe we should adopt HTML reports similar to other python tools. 

e.g. (`pytest-html`, the `coverage html` command, ect...)

Though I do like the simple styling of PIT :P

s-kovacevic:
Exactly.
I am not suggesting copy and paste of PIT's report CSS @boxed, just throwing this idea of HTML report out there.

Thats why I wanted to hear some feedback. If this can help with some use case, even better, maybe good intergration with CI/CD tools, that would certanly be beneficial in my opinion.

boxed:
Does anyone want to give this 20 minutes? Or should we wait till next time I have insomnia and I whip something up? :)

boxed:
I have merged my pretty rough html reports and made a release.

PR #72, opened 2019-01-14
HTML report

boxed:
@s-kovacevic @nklapste check this rough draft out. Maybe @ulope is also interested?

I've attached a sample output from tri.declarative:

[html.zip](https://github.com/boxed/mutmut/files/2754678/html.zip)

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -22,7 +22,7 @@ from glob2 import glob
 from mutmut import mutate_file, Context, list_mutations, __version__, \
     BAD_TIMEOUT, \
     OK_SUSPICIOUS, BAD_SURVIVED, OK_KILLED, UNTESTED
-from mutmut.cache import hash_of_tests
+from mutmut.cache import hash_of_tests, create_html_report
 from mutmut.cache import register_mutants, update_mutant_status, \
     print_result_cache, cached_mutation_status, \
     filename_and_mutation_id_from_pk, cached_test_time, set_cached_test_time, \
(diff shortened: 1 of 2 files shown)


boxed:
(This references issue #61)

nklapste:
I definitely like the ease of use with these html reports!

This maybe for a future PR, but, adding a line highlighting feature similar to how PIT reports would be useful. They allow for quick notation of what lines are causing mutation failures, thus, allowing a developer to quickly add the `# pragma: no mutate` as needed. As shown in the reports you generated only a small section of code (lines 764 to 864 posed problems), having a highlight of these issues could prove beneficial in giving a better visualization of failing code.

boxed:
Highlighting the code seems like it's not really useful if you can't get at the mutants though. How about showing the full source code with color highlighting for lines that contain mutants and then allow you to click on those lines and that expands to the information in the current report (i.e. the diff and mutation id)?

boxed:
@s-kovacevic @nklapste @sobolevn So what should we do at this point? Do you guys think my last idea for a different HTML output is better and this current version should be scrapped or should we have some different output options? Or maybe something else?

s-kovacevic:
I think for now it is good to go, it is certanly better than nothing. However I think we should aim to improve this in future. I think this is important to improve mutmut's CI/CD capabilities. Also moving to html lib or jinja would make the code readable and easier to work with instead of inline html strings.

The original issue should probably remain open for any further discussion on how to improve HTML reports.

fipa:
Hi, I'm starting with mutmut, moving from Java (where we have pitest) to Python and this is the one thing I'm missing. Is there any plans on moving on with this feature?

Thanks!

boxed:
I guess. The problem i have is that I don't personally care and don't see myself using this feature so it's hard for me to know what would be a good format and design. Maybe you can provide some insight or even help design it?

boxed:
Hmm... Reading the last comment it seems like I should merge this and make a new release and then wait for feedback. Is that a good plan for you?

fipa:
@boxed that sounds excellent. I will use it so it won't be a problem for me to give you any feedback on the design, or maybe even PR myself.

boxed:
This is merged and released. Give it a try! 

Note that it's super slow to generate the report. This is a general limitation in how mutmut generates mutants. It could be improved by mutating the AST, then putting it back to the state it was. Currently mutmut reparses the entire AST, which is slow (but faster than a deepcopy of the AST!).

PR #74, opened 2019-01-17
Improve `__init__.py` tests, remove coverage for windows specific code

s-kovacevic:


boxed:
Could we maybe get the coverage data from all the build stages and merge them? That would be much nicer since we can remove most of the coverage exclusions and get real numbers.

nklapste:
@s-kovacevic a return code of 4 for `CliRunner().invoke(climain, ['run', '--paths-to-mutate=foo.py', "--test-time-base=5.0"], catch_exceptions=False)` indicates that a mutant timed out (i.e. `BAD_TIMEOUT` mutant status). 

I think this is just the travis-ci windows environment being slower than the Linux ones (linux py3.7 can complete tests in 35.80sec vs windows py3.7 in 48.01sec).

We might have to increase the `--test-time-base` to accommodate. For example:
`CliRunner().invoke(climain, ['run', '--paths-to-mutate=foo.py', "--test-time-base=10.0"], catch_exceptions=False)`  might resolve this.

Issue #84, opened 2019-02-16
pytest output plugin


sobolevn:
Hi, @boxed! 

I am running `mutmut` on my code and it works fine. 
But, mine goal is to literally see which tests are good and which tests are bad.

So, what I suggest: let's create a `pytest` plugin that will report and save failed tests into some sort of a database/file. We will have a list of failed tests and then we can find useless tests as a result.

What do you think about this?

boxed:
Interesting! I agree that tests that never fail during mutation testing are very likely bogus. If you run pytest with the -x flag we could also find redundant tests. 

I'm hesitant to build functionality that depends on pytest. How about a feature to add a post and pre command to mutation testing? In your case you could use it to run pytest with the xml output format and in the post step run some command to collect the xml and store it with a unique name. Then you need to write a bit of code to parse all those xml files to get the data you want. 

How does this sound?

sobolevn:
Sounds like a plan!

boxed:
Check out the branch I've created with this feature. It's pretty straight forward code wise, just need you to try it out and see if it does what you want.

boxed:
Well I've merged the PR, so I'll close this. Hope this works for you!

PR #85, opened 2019-02-20
Post and pre hooks for the mutation step

boxed:
This is a prototype of solving the problem in #84

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -35,14 +35,22 @@ if sys.version_info < (3, 0):   # pragma: no cover (python 2 specific)
     orig_print = print
 
     def print(x='', **kwargs):
+        # noinspection PyUnresolvedReferences
         x = x.decode("utf-8")
         orig_print(x.encode("utf-8"), **kwargs)
 
+    # noinspection PyShadowingBuiltins
     class TimeoutError(OSError):
         """Defining TimeoutError for Python 2 compatibility"""
+
+    # noinspection PyShadowingBuiltins
+    class FileNotFoundError(OSError):
+        """Defining FileNotFoundError for Python 2 compatibility"""
 else:
     # noinspection PyUnresolvedReferences,PyCompatibility
     from configparser import ConfigParser, NoOptionError, NoSectionError
+
+    # noinspection PyShadowingBuiltins
     TimeoutError = TimeoutError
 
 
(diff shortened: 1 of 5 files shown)


mutmut/__main__.py
```diff
@@ -506,6 +523,9 @@ def run_mutation(config, filename, mutation_id):
     if cached_status != UNTESTED:
         return cached_status
 
+    if config.pre_mutation:
+        subprocess.check_output(config.pre_mutation, shell=True)
```
sobolevn: What will happen if any of the pre/post mutation process fails?

boxed: Then it terminates mutmut. I think that makes sense as a default, otherwise it's pretty hard to debug the post/pre hooks.

sobolevn: I guess the exception log from failing processes must be shown to the end user.

boxed: Yea, that's what you get by default. In fact you'll get all the stdout and stderr of the called process so your pre/post hooks should probably be silent in the normal case to not destroy the output of mutmut.

sobolevn: Awesome! 👍

mutmut/__main__.py
```diff
@@ -533,6 +553,9 @@ def run_mutation(config, filename, mutation_id):
     finally:
         move(filename + '.bak', filename)
 
+        if config.post_mutation:
```
nklapste: we should add a simple integration test that ads some coverage over these statements just for good measure.

Maybe something like:

```python
def test_post_mutation_hook(filesystem):
    result = CliRunner().invoke(climain, ['run', '--paths-to-mutate=foo.py', "--test-time-base=15.0", """--post-mutation='python -c "print(\'post mutation stub\')""""], catch_exceptions=False)
    print(repr(result.output))
    assert result.exit_code == 0
    assert "post mutation stub" in result.output
```

boxed: Good idea. I modified the test a little bit from what you wrote, and it did show a bug! I swallowed the output which would have been super annoying.

tests/test_main.py
```diff
@@ -339,3 +339,19 @@ def test_use_coverage(capsys, filesystem):
...
+def test_pre_and_post_mutation_hook(filesystem):
+    result = CliRunner().invoke(
+        climain, [
+            'run',
+            '--paths-to-mutate=foo.py',
+            "--test-time-base=15.0",
+            "--pre-mutation=echo pre mutation stub",
+            "--post-mutation=echo post mutation stub",
+        ], catch_exceptions=False)
+    print(result.output)
```
sobolevn: Looks like this `print` is not required

boxed: It isn't, but when the test fails it's super nice to have.

PR #89, opened 2019-02-27
Fixes #87

### LINKED ISSUE #87: mutmut does not mutate function internals
I have a simple function:

```python
def add(first: float, second: float) -> float:
     return first + second
```

And a test that ensures that it is working:

```python
# import add

def test_add():
     assert add(0, 0) == 0
```

And it works with `pytest`.

But, this should fail with `mutmut`. I expect it to change `+` -> `-`. Here's the logic that does it: https://github.com/boxed/mutmut/blob/master/mutmut/__init__.py#L293

And I expect this mutant to pass. But, nothing really happens. 
I have create a reproduction repo for you: https://github.com/sobolevn/heisenbug-2019

Here's what it outputs without cache:

```
» mutmut run

- Mutation testing starting -

These are the steps:
1. A full test suite run will be made to make sure we
   can run the tests successfully and we know how long
   it takes (to detect infinite loops for example)
2. Mutants will be generated and checked

Mutants are written to the cache in the .mutmut-cache
directory. Print found mutants with `mutmut results`.

Legend for output:
🎉 Killed mutants. The goal is for everything to end up in this bucket.
⏰ Timeout. Test suite took 10 times as long as the baseline so were killed.
🤔 Suspicious. Tests took a long time, but not long enough to be fatal.
🙁 Survived. This means your tests needs to be expanded.

mutmut cache is out of date, clearing it...
1. Running tests without mutations
⠸ Running... Done

2. Checking mutants
⠴ 0/0  🎉 0  ⏰ 0  🤔 0  🙁 0%
```

sobolevn:
I have test it locally, but I edited this line via github.

I will add a testcase for it later.

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -557,7 +557,7 @@ def mutate_list_of_nodes(node, context):
     for child_node in node.children:
 
         if child_node.type == 'operator' and child_node.value == '->':
-            return
+            continue
 
         mutate_node(child_node, context=context)
 


boxed:
Hah, well that's clearly an improvement, but it does break some other things unfortunately. We need to add a rule to not mutate the return type annotation.

sobolevn:
@boxed sure. I will work on this tomorrow.

boxed:
@sobolevn great! Try to use the ASTPattern class. I want it to be easy to use, so any problems you'll have are very much appreciated if you can share them :P

I've also invited you to be a collaborator on mutmut now.

sobolevn:
@boxed awesome! Thanks for inviting me 😊 

WIP

boxed:
Ping? This seems pretty important, so if you don't have time to work on it, just say the word and I'll take over.

sobolevn:
@boxed yes, please help me with this one. Currently I am quite busy with the business project. 😞

boxed:
Blech, this was much more icky to fix than I'd hoped. I'll make a totally new change :(

sobolevn:
@boxed thanks for your help!

Btw, I am making a talk about `mutmut`, I have started to work on examples: https://github.com/sobolevn/heisenbug-2019 It is an early alpha. Maybe you have some feedback about it? I would appreciate that.

boxed:
I'll take a look!

Issue #43, opened 2018-11-26
Mutate only changes from a patch


blueyed:
I think a very useful feature would be to only mutate code that has been changed in a patch (e.g. what is in the Git index etc).

This way you could run mutmut before/after committing changes to check if everything is covered with regard to mutations in a new commit (which not enforcing checks on the whole code base).

boxed:
Depends on the purpose I guess... if you have a zero mutant project where you don't want to introduce new mutants you can already do this if you just save the .mutmut-cache file and run the changes based on it. 

What is the use case you're thinking about?

blueyed:
The use case is having a non-zero mutant project, where you want to start eliminating mutants based on new changes.
Similar to covering a PR/commit completely with tests (i.e. diff coverage should be >= X), I would like to ensure that there are no (new) mutants.
For this mutations should get applied only to the new code (changed in the patch/diff).

boxed:
Yea ok. In this case it should be trivial since the coverage stuff is basically this, if one can just extract the lines from the patch. Do you know an easy way to do that?

boxed:
This has now been implemented and merged!

PR #90, opened 2019-03-01
Support for mutating only lines specified by a patch file

boxed:
This is the implementation for #43

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -605,7 +605,6 @@ def run_mutation_tests(config, mutations_by_file):
 
 def read_coverage_data():
     """
-    :type use_coverage: bool
     :rtype: CoverageData or None
     """
     print('Using coverage data from .coverage file')
@@ -622,7 +621,7 @@ def read_patch_data(patch_file_path):
         # noinspection PyPackageRequirements
         import whatthepatch
     except ImportError:
-        print('The --use-patch feature requires the whatthepatch library. Run "pip install whatthepatch"')
+        print('The --use-patch feature requires the whatthepatch library. Run "pip install whatthepatch"', file=sys.stderr)
         raise
     with open(patch_file_path) as f:
         diffs = whatthepatch.parse_patch(f.read())
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -201,17 +201,20 @@ DEFAULT_TESTS_DIR = 'tests/:test/'
 @click.option('--untested-policy', type=click.Choice(['ignore', 'skipped', 'error', 'failure']), default='ignore')
 @click.option('--pre-mutation')
 @click.option('--post-mutation')
+@click.option('--use-patch-file', help='Only mutate lines added/changed in the given patch file')
 @config_from_setup_cfg(
     dict_synonyms='',
     runner='python -m pytest -x',
     tests_dir=DEFAULT_TESTS_DIR,
     pre_mutation=None,
     post_mutation=None,
+    use_patch_file=None,
 )
 def climain(command, argument, paths_to_mutate, backup, runner, tests_dir,
             test_time_multiplier, test_time_base,
             swallow_output, use_coverage, dict_synonyms, cache_only, version,
-            suspicious_policy, untested_policy, pre_mutation, post_mutation):
+            suspicious_policy, untested_policy, pre_mutation, post_mutation,
+            use_patch_file):
     """
 commands:\n
     run [mutation id]\n
@@ -233,13 +236,14 @@ commands:\n
                   tests_dir, test_time_multiplier, test_time_base,
                   swallow_output, use_coverage, dict_synonyms, cache_only,
                   version, suspicious_policy, untested_policy, pre_mutation,
-                  post_mutation))
+                  post_mutation, use_patch_file))
 
 
 def main(command, argument, paths_to_mutate, backup, runner, tests_dir,
          test_time_multiplier, test_time_base,
          swallow_output, use_coverage, dict_synonyms, cache_only, version,
-         suspicious_policy, untested_policy, pre_mutation, post_mutation):
+         suspicious_policy, untested_policy, pre_mutation, post_mutation,
+         use_patch_file):
     """return exit code, after performing an mutation test run.
 
     :return: the exit code from executing the mutation tests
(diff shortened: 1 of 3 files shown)


boxed:
I like this way of using an optional feature better than adding another dependency to mutmut. I think it would be nice if we moved junitxml and coverage to do this too, so "pip install mutmut" becomes smaller and faster when most people won't use those features anyway.

What do you guys think?

test_requirements.txt
```diff
@@ -1,4 +1,5 @@
 pytest>=2.8.7,<4.0.0
 pytest-cov<=2.6.0
 mock==2.0.0
-coverage
\ No newline at end of file
+coverage
```
nklapste: Maybe we should move these requirements into the `setup.py` `setup()`'s `extras_require` kwarg. They seem like tack-on extra functionality / features so they meet the use case of that kwarg.

It could be implemented something like this:
```python
setup(
    # other code....
    extras_require={
        "patch_filtering": [
            "whatthepatch",
        ],
        "coverage_filtering": [
            "coverage",
        ],
    },
)
```
That way if someone wants the enable the coverage filtering they could install it via pip with:

```shell
pip install mutmut["coverage_filtering"]
```

And if they wanted to enable patch filtering they could install via pip with:
```shell
pip install mutmut["patch_filtering"]
```

This should help limit core dependencies and lower the chance of dependency conflict with other peoples projects.

test_requirements.txt
```diff
@@ -1,4 +1,5 @@
...
+whatthepatch
```
nklapste: We should lock this requirement at least on the API level `>=0.0.5,<1.0.0`.

mutmut/__main__.py
```diff
@@ -587,20 +602,34 @@ def run_mutation_tests(config, mutations_by_file):
         run_mutation_tests_for_file(config, file_to_mutate, mutations)
 
 
-def read_coverage_data(use_coverage):
+def read_coverage_data():
     """
     :type use_coverage: bool
```
nklapste: remove these docstrings now that they are irrelevant

nklapste: Remove these docstrings now that they are irrelevant.

mutmut/__main__.py
```diff
@@ -329,27 +336,35 @@ def main(command, argument, paths_to_mutate, backup, runner, tests_dir,
     if using_testmon:
         copy('.testmondata', '.testmondata-initial')
 
-    if not use_coverage:
-        def _exclude(context):
-            del context
-            return False
-    else:
+    if use_coverage or use_patch_file:
```
nklapste: It might be cleaner to separate these conditions into separate `if` statements and deal with the code duplication by refactoring it out into functions.

Otherwise, just allow the duplication and separate the `if` statements anyways.
e.g.
```python
if use_coverage:
    # use coverage setup
elif use_patch_file:
    # use patch file
```

Having them together makes the branching somewhat hard to understand.

mutmut/__main__.py
```diff
@@ -201,17 +201,20 @@ def print_progress(self):
 @click.option('--untested-policy', type=click.Choice(['ignore', 'skipped', 'error', 'failure']), default='ignore')
 @click.option('--pre-mutation')
 @click.option('--post-mutation')
+@click.option('--use-patch-file', help='Only mutate lines added/changed in the given patch file')
```
nklapste: Should group this next to the `@click.option('--use-coverage', is_flag=True, default=False)` line for clarity.

mutmut/__main__.py
```diff
@@ -587,20 +602,34 @@ def run_mutation_tests(config, mutations_by_file):
...
-        return None
+    print('Using coverage data from .coverage file')
+    # noinspection PyPackageRequirements,PyUnresolvedReferences
+    from coverage import Coverage
+    cov = Coverage('.coverage')
+    cov.load()
+    return cov.get_data()
+
+
+def read_patch_data(use_patch_file):
```
nklapste: We should rename the arg `use_patch_file` to `patch_file_path` so that it is more self descriptive.

mutmut/__main__.py
```diff
@@ -587,20 +602,34 @@ def run_mutation_tests(config, mutations_by_file):
...
+    print('Using patch data from ' + use_patch_file)
+    try:
+        # noinspection PyPackageRequirements
+        import whatthepatch
+    except ImportError:
+        print('The --use-patch feature requires the whatthepatch library. Run "pip install whatthepatch"')
+        return -1
```
nklapste: We should just re-raise the exception. 
```python

 try:
    # noinspection PyPackageRequirements
    import whatthepatch
except ImportError:
    print('The --use-patch feature requires the whatthepatch library. Run "pip install whatthepatch"')
    raise
```


On an python exception fully raising and terminating execution a return code of `1` should be set. This conforms to our exit code standard of 1 being some error within the execution of mutmut. Also the user then gets the full traceback on the import error, this could be useful for debugging if the import error is a fault on the users end (e.g. corrupt python environment).

mutmut/__main__.py
```diff
@@ -587,20 +602,34 @@ def run_mutation_tests(config, mutations_by_file):
...
+    return cov.get_data()
+
+
+def read_patch_data(use_patch_file):
+    print('Using patch data from ' + use_patch_file)
+    try:
+        # noinspection PyPackageRequirements
+        import whatthepatch
+    except ImportError:
+        print('The --use-patch feature requires the whatthepatch library. Run "pip install whatthepatch"')
```
nklapste: We should probably print to stderr
```python
sys.stderr.write('The --use-patch feature requires the whatthepatch library. Run "pip install whatthepatch"\n')
```

test_requirements.txt
boxed: Cool. Didn't know about that. I'll punt on this for now and we'll do this in a new PR after this lands.

mutmut/__main__.py
boxed: I see your point but it's not obvious to me how to improve it. I'll add a comment do explain why they are grouped together for now.

test_requirements.txt
s-kovacevic: Yeah, that would be cool.

mutmut/__main__.py
boxed: print('foo', file=sys.stderr)

PR #107, opened 2019-03-27
Rename long option for `-s`

blueyed:
It actually does not swallow output, so `--no-capture` is a better name.

Based on that pytest has `--capture=no` for `-s`.
Maybe something positive like `--keep-output` or `--display-output` would be better?


boxed:
Aha. Yea then I think we should follow pytest.

bk521234 (approved):

l0b0 (approved):

l0b0:
@blueyed Not sure if you're aware, but this needs a rebase.

PR #116, opened 2019-04-22
Switch setup.py to 'console_scripts' from 'scripts'

### LINKED ISSUE #115: CLI "mutmut ..." commands don't work on Windows
The simple `$ mutmut ...` invocations work fine on my Debian stretch box, but don't work at all on Windows, giving 'command not found' type errors. `python -m mutmut ...` works fine, though.

Is there any reason `mutmut` isn't using the [`console_scripts` setuptools entry point](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point) for the CLI invocations? If not, I'd like to put together a PR to set that up...AFAIK it *should* work uniformly on all platforms. 

Alternatively, I'd propose creating a new `mutmut.bat`, adapted from & to put alongside the current `/bin/mutmut`, to enable streamlined Win CLI invocations.

bskinn:
Closes #115.

This configuration of `setup.py` results in a `mutmut` CLI script that works for me both on Windows 10 and on Debian stretch.

### DIFF
--- a/bin/mutmut
+++ b/bin/mutmut
@@ -1,5 +1,0 @@
-#!/usr/bin/env python
-
-if __name__ == '__main__':
-    from mutmut.__main__ import climain
-    climain()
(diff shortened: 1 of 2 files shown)


boxed:
Thanks!

Issue #117, opened 2019-05-01
Should surviving suspicious mutants count towards total mutants?


jazwiecki:
In `__main__.py/run_mutation()`, both killed and surviving mutants count as "OK_SUSPICIOUS". Before looking through the code, I thought that the "suspicious" classification was meant to identify killed mutants that took a suspiciously long time to die, indicating that there might be a mutation with performance implications that should be checked even if it passed. While running `mutmut`, I was surprised to see the number of "Suspicious" mutants incremented with _both_ passes and fails. For the purposes of mutation testing, I think that the "Suspicious" count should represent either killed or surviving mutants, not both. Here's [the relevant part of the code](https://github.com/boxed/mutmut/blob/589c4bf7286ab59a22c428662c4ac8dfef521631/mutmut/__main__.py#L577-L593):

```
def run_mutation(config, filename, mutation_id):

        [...]

        try:
            survived = tests_pass(config)
        except TimeoutError:
            context.config.surviving_mutants_timeout += 1
            return BAD_TIMEOUT

        time_elapsed = time() - start
        if time_elapsed > config.test_time_base + (config.baseline_time_elapsed * config.test_time_multipler):
            config.suspicious_mutants += 1
            return OK_SUSPICIOUS

        if survived:
            context.config.surviving_mutants += 1
            return BAD_SURVIVED
        else:
            context.config.killed_mutants += 1
            return OK_KILLED
```

I'm happy to write a PR to split up suspicious surviving and killed mutant counts, but I wanted to confirm whether or not that work would be in keeping with the design of mutmut. My first though would be to _only_ mark a mutant as suspicious if `not survived`, in addition to the existing run time threshold check. Let me know what you think.

boxed:
I totally agree, it's clearly a bug. 

A PR would be great!

jazwiecki:
OK, PR here: https://github.com/boxed/mutmut/pull/118

Didn't have time to figure out how the tests were set up (sorry!!!) but if anyone has a tip on where/how to make sure this is covered I can do the work.

PR #118, opened 2019-05-02
don't count surviving mutants as suspicious

jazwiecki:
re: https://github.com/boxed/mutmut/issues/117, not sure how to add tests to this but happy to do it if anyone has advice


boxed:
It's fine without a test I think. At least for now. Thanks for this!

PR #129, opened 2019-06-14
Feature/multiple primitive mutations

nklapste:
Adds the ability for operator mutations (and possibly other primitive mutations) to have multiple alternatives.

This should be able to address the request of additional mutations for assignment operators as noted in #124.

For example the python statement:

```python
x += 2
```
could now be mutated to both the originally supported mutmut mutation:
```python
x -= 2
```
and a new alternative mutation:
```python
x = 2
```

PR #131, opened 2019-07-04
Feature/remove tri declaritive

nklapste:
Optional PR to be applied on top of #129 

Removes the requirement of `tri.declarative` and potential future issues with the library.


boxed:
Well, tri.declarative is a library I am one of two maintainers on so the issue created (the module renaming) was of my own making so not really something I worry about more of :P

Buy yea, just using evaluate here where we always want to just call a function seems silly.

nklapste:
Adding the removal of `tri.declarative` into #129

Issue #103, opened 2019-03-16
Failed to abort an infinite loop


nedbat:
(This might be related to my unusual settings described in #102)

mutmut got stuck on a mutation, and didn't time it out.  Reproduction:
```
$ mktmpenv -p python3.7
Running virtualenv with interpreter /usr/local/bin/python3.7
Using base prefix '/usr/local/pythonz/pythons/CPython-3.7.2'
/usr/local/pythonz/pythons/CPython-2.7.14/lib/python2.7/site-packages/virtualenv.py:1047: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
New python executable in /usr/local/virtualenvs/tmp-b4da94a5634daf06/bin/python3.7
Also creating executable in /usr/local/virtualenvs/tmp-b4da94a5634daf06/bin/python
Installing setuptools, pip, wheel...done.
Requirement already up-to-date: pip in /usr/local/virtualenvs/tmp-b4da94a5634daf06/lib/python3.7/site-packages (19.0.3)
Requirement already up-to-date: setuptools in /usr/local/virtualenvs/tmp-b4da94a5634daf06/lib/python3.7/site-packages (40.8.0)
This is a temporary environment. It will be deleted when you run 'deactivate'.

$ git clone git@github.com:nedbat/cog.git
Cloning into 'cog'...
remote: Enumerating objects: 657, done.
remote: Counting objects: 100% (657/657), done.
remote: Compressing objects: 100% (230/230), done.
remote: Total 657 (delta 458), reused 619 (delta 421), pack-reused 0
Receiving objects: 100% (657/657), 172.29 KiB | 574.00 KiB/s, done.
Resolving deltas: 100% (458/458), done.

$ cd cog

$ git checkout c1f605d7aaf3750568b7e7a5d784c081fff0fc97
Note: checking out 'c1f605d7aaf3750568b7e7a5d784c081fff0fc97'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at c1f605d mutmut settings

$ pip install tox==3.7.0
Collecting tox==3.7.0
[...]

PR #132, opened 2019-07-11
Feature/drop python27

nklapste:
Dropping Python27 and adopting more Python3-isms

This also paves the way for using `asyncio` for multiprocessed output streaming as mentioned in #103.


nklapste:
@boxed Shippable keeps attempting a Python2.7 is that from configurations defined on your end?

boxed:
Shippable is supposed to just use the travis config, so not that I'm aware of.

nklapste:
Besides Shippable failing due to configuration that is outside of my control (and/or requiring separating our build scripts). This PR should be ready for a quick review.

Any improvements that could be onset by adopting only python3 will likely be added in future PR. This more or less removes any python2.7 fixes and mentions.

I would recommend choosing either _Travis-ci_ or Shippable maintaining both build scripts is a waste of resources.

boxed:
Agreed. I set up shippable when the news of Travis development being gutted got out. But let's drop it until we need it.

boxed:
The changes to .format() seems a bit silly to me. It makes more sense to change to f-strings (and require 3.6).

blueyed:
> The changes to .format() seems a bit silly to me.

It could also just keep the percent formatting, no?
(just out of interest)

nklapste:
> 
> 
> > The changes to .format() seems a bit silly to me.
> 
> It could also just keep the percent formatting, no?
> (just out of interest)

I've had a few issues before when debugging mutmut where the `%` formatted strings were either causing premature Exceptions or getting in the way due to typing.

I like the fact that `format` goes pretty far to try and serialize what your giving it and is less picky than `%` formatting strings on what is given (or not given).

It would be best if we adopted one form of string formatting just for code cleanliness. 

If we were to push to only support 3.6+ I would too recommend moving to f-strings.

PR #137, opened 2019-08-12
WIP: `asyncio` Usage

nklapste:
Using the python `asyncio` library to replace the current implementation of `popen_streaming_output`.

This currently does not support Python 3.4 due to the usage of the keywords `async` and `await`.


boxed:
Does this actually _work_ though? I mean, if you write a program ```while True: print('.', end='')``` and run it through this system, does it get the input streamed? On linux, windows and macOS?

nklapste:
> 
> 
> Does this actually _work_ though? I mean, if you write a program `while True: print('.', end='')` and run it through this system, does it get the input streamed? On linux, windows and macOS?

I'm assuming you meant say output instead of the input. 

The following test was run over Linux and Windows (I removed the ``bash`` call in Windows). Both streamed their stdout and stderr without blocking.

```python
from mutmut.__main__ import popen_streaming_output


def feedback(line):
    print(line.decode("utf-8"), end="")


popen_streaming_output("""bash -c 'python -c \"for i in range(1000000000): print(1, end=str( ));\"'""", callback=feedback, timeout=5)
```

boxed:
Input and output depends on perspective: it's the test output but it's mutmuts input :P

boxed:
:( bad news. I tried the test above, but it doesn't stream the data on my macOS system. I just get all the ones at the end.

nklapste:
> 
> 
> :( bad news. I tried the test above, but it doesn't stream the data on my macOS system. I just get all the ones at the end.

Are you using a stale version of this branch? The newest version calls the printing callback on each byte received (via calling `read(1)` instead of `readline()`). Thus, it should stream each incoming 1 character. 

Otherwise, could you see if in a standard `mutmut run` command the spinner behaves normally on macOS? Maybe my test is flawed for all systems.

I'll try and setup a macOS vm for testing at some point. But, I would be very surprised that behavior shared on both Linux and Windows is not present in some form on macOS.

boxed:
I got the branch just before trying and I posted a few minutes after testing, so I don't see how I could have gotten a stale branch?

nklapste:
Closed replaced with #199

PR #142, opened 2019-08-29
Bugfix/compatibility with whatthepatch==0.0.6

nklapste:


nklapste:
outdated with new master c:

PR #145, opened 2019-08-29
WIP: Different workers

boxed:
This is a big architecture change so I'd like to get some feedback!

- mutmut always spawns a process that actually performs the mutations, it communicates with this via a subprocess channel
- there are now two worker types:
  - separate_process: this is like the old thing pretty much, except it's being executed from another process (yea, this becomes a bit silly but it's simpler when combined with the next one worker type)
   - same_process: this directly executes in the same process and performs some module unloading tricks to make this work.

Advantages:

- The same_process worker type is ~2x faster for small test suites like tri.struct and tri.declarative!
- I think this should make it pretty simple to implement multiple parallel workers in the future.

TODO:
- timeout for same_process

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
+2.0.0 (2019-08-29)
+~~~~~~~~~~~~~~~~~~
+
+* Removed support for testmon. This was an experimental feature that didn't give any significant speedups unfortunately.
+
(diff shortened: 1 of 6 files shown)


boxed:
@nklapste

boxed:
Gave this a good try on tri.declarative, but it looks like it leaves mutants in place after it's done sometimes, and ctrl+c leaves mutants plus .bak-files :(

boxed:
Never got this to work properly :(

PR #147, opened 2019-09-04
test_requirements.txt: Allow more recent versions

jayvdb:


jayvdb:
More a POC - I expect that the maintainers would want more refined pins than I have provided here.  I have not researched the reasons for the existing pins.  I am happy to adjust them, or the maintainers can discard this and do it themselves if that is easier.

boxed:
I think we need to lock whatthepatch to the exact version we've tested because it's pre 1.0 so it ships backwards incompatible changes.

jayvdb:
I didnt modify that pin, except for the EOL which was unintentional.  Anyways, happy to adjust it to ==0.0.6.  I guess that means you are happy with the rest of the changes.

boxed:
Yes, I'm fine with this change! Just lock down whatthepatch. I locked pytest before because I was upset after 4.0 broke a bunch of my stuff, but it's better to accept this pain and move forward!

jayvdb:
I've also updated `setup.py` and `__main__.py` to make that pinning a bit more holistic.

boxed:
Nice.

Issue #76, opened 2019-01-23
No mutations when running with --use-coverage


Vlad-Shcherbina:
Steps to reproduce:
1. In an empty directory, create the file `example_test.py` with the following content:
```python
def test_stuff():
    assert True
```
2. Run `pytest --cov` to generate `.coverage` file.
3. Run `mutmut run --paths-to-mutate=example_test.py --use-coverage`.

Expected result: 1 killed mutant (I tried running mutmut without `--use-coverage` and there was indeed 1 mutant; and `coverage html` shows that that line is covered).

Actual result: no mutants whatsoever.
```
1. Running tests without mutations
⠇ Running... Done
Using coverage data from .coverage file

2. Checking mutants
⠏ 0/0  🎉 0  ⏰ 0  🤔 0  🙁 0
```

--------

Version info:
```
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)] on win32
pytest==4.1.1
pytest-cov==2.6.1
mutmut==1.2.0
```

boxed:
I believe this is the same issue as https://github.com/boxed/mutmut/pull/69 as I can reproduce it with the released version but not with the latest master. I'm pushing a release.

The reproduction scenario you list above is a rather pathological case though, as the mutant is killed because the test itself is modified :P

boxed:
A 1.3.0 release is on its way. The entire build and release takes ~10 minutes.

Vlad-Shcherbina:
Thank you, I confirm it works with mutmut 1.3.0.

I know it's pathological, I was going for the simplest way to reproduce.

HitLuca:
mutmut still doesn't have any mutants when the `--use-coverage` flag is used. Working with PyCharm and the IDE coverage sourced from .coverage correctly shows covered statements in the various files.

Without the `--use-coverage` flag all mutants are correctly found

Command:
`mutmut run --paths-to-mutate file.py --tests-dir tests/`

Version info:
```
mutmut==1.5.1
pytest==5.1.1
pytest-cov==2.7.1
```

HitLuca:
[...]

PR #149, opened 2019-09-05
Check for non existent filepaths when using coverage data

HitLuca:
fixed issue #76 

Added a simple check after loading the .coverage data: if a path doesn't exist raise an exception, ensuring that the .coverage file is usable.

Especially needed when generating coverage with PyCharm


HitLuca:
Loving Windows avoiding my regex

boxed:
Thanks!

HitLuca:
No worries, happy to help!

Luca

Issue #127, opened 2019-06-03
Bogus mutation: "in" -> "not in" inside generator expression


boxed:
Extremely similar to https://github.com/boxed/mutmut/issues/20

PR #158, opened 2019-10-19
Enhancement : 'in' in generators do not mutate

roxanebellot:
Fix related to issue https://github.com/boxed/mutmut/issues/127


boxed:
Thanks! I've sent an invite to give you merge/commit access.

Issue #172, opened 2020-05-20
Github Actions as a CI?


atugushev:
I've noticed that the current TravisCI config looks pretty neglected:
- only Linux 
- only Python 3.7
- no lint job
- no codecov integration (i see codecov mentions, but it doesn't run)

I was tempted to improve the config at first, but since I personally prefer Github Actions I've decided to ask, there is might be an interest in Github Actions CI integration? It has 20 parallel jobs, works with Windows/Linux/macOS, supports all popular kinds of python, and the package can be also deployed to PyPI using a GitHub action. I'd love to work on it if there is a "yes". Later it can replace altogether the Travis.

If it's a "nay", I'd like to improve the Travis config anyway.

boxed:
The reason it's so cut down is because of the new execution model. I sacrificed windows compatibility for big improvements on macOS and Linux. I do want windows support I just don't have a windows machine to test on and pushing random ideas to Travis to see if they worked wasn't working. 

So windows support seems like a separate thing. 

I am curious about github actions so I'd like to see how that could work. Maybe you have an example? Otherwise I don't really have an opinion except that ultimately it falls on me to maintain it so it's nice if it's the same as other projects I also have to maintain.

atugushev:
> Maybe you have an example?

I'll prepare a PR with a GA workflows then. Let's see how it goes then. In the meantime, I'll send a PR to add a `lint` job to Travis CI.

boxed:
If you can make the lint just run on PRs and just warn then yes please. But failing the build? Hell no. I hate that.

atugushev:
Thanks for the reply!

PR #173, opened 2020-05-21
Revisit Travis CI config

atugushev:
- add py35, py36, py38, py39 jobs
- add lint job (allowed to fail)
- run tests with coverage and fix coverage report sending to codecov

Refs #172.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,37 @@
 language: python
+python:
+  - 3.6
+  - 3.7
+  - 3.8
+  - 3.9-dev
+
+env:
+  - TOXENV=coverage
 
 stages:
   - lint
   - test
-  - build
-  - deploy
+  - name: deploy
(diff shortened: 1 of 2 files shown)


boxed:
That looks a lot cleaner than I had expected! Great stuff!

boxed:
Go ahead and merge if you're ready.

atugushev:
Cool! Thanks @boxed  for the feedback!

boxed:
3.8 fails bc of lint and 3.9 because pony doesn't allow it. We should report to pony at least!

atugushev:
>  3.8 fails bc of lint 
Yeah, I intentionally didn't fix lint error to demonstrate that it allowed to fail. Will be fixed in a following up PR.

> 3.9 because pony doesn't allow it. We should report to pony at least!
Right!

boxed:
I've reported to pony.

PR #184, opened 2020-10-08
Fixed the total mutants and killed column in html report

peterfortuin:
The total number of mutants column and the number killed column should be switched if you look at the table headers.


boxed:
Oops! Thanks for this fix!

peterfortuin:
Thats fast! Nice work!

PR #192, opened 2020-12-11
Feature/simple output

### LINKED ISSUE #144: UTF-8 emoji icons fail to display in a Pycharm 2019.2.1 terminal in windows
This was created on a windows 10 environment with Pycharm 2019.2.1.

Running mutmut in the Pycharm native terminal fails to render the emojis indicating killed 🎉, suspicious 🤔, and surviving 🙁 mutants.

For example:
```console 
(venv) >mutmut run --paths-to-mutate my_project

- Mutation testing starting -

These are the steps:
1. A full test suite run will be made to make sure we
   can run the tests successfully and we know how long
   it takes (to detect infinite loops for example)
2. Mutants will be generated and checked

Results are stored in .mutmut-cache.
Print found mutants with `mutmut results`.

Legend for output:
� Killed mutants.   The goal is for everything to end up in this bucket.
⏰ Timeout.          Test suite took 10 times as long as the baseline so were killed.
� Suspicious.       Tests took a long time, but not long enough to be fatal.
� Survived.         This means your tests needs to be expanded.

mutmut cache is out of date, clearing it...
1. Running tests without mutations
...
```

Maybe we should simply adopt a ASCII text based notation?

nklapste:
Fixes #144

Adds the option `--simple-output` to swap the usage of emojis to plain text alternatives.

This resolves cases where the terminal environment fails to properly print/support emojis!

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -910,8 +910,9 @@ def guess_paths_to_mutate():
 
 
 class Progress(object):
-    def __init__(self, total):
+    def __init__(self, total, output_legend):
         self.total = total
+        self.output_legend = output_legend
         self.progress = 0
         self.skipped = 0
         self.killed_mutants = 0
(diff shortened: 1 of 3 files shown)


boxed:
Thanks!

PR #197, opened 2021-01-13
Add python_requires metadata

hukkinj1:
This metadata helps pip not install versions of `mutmut` that don't support the interpreter in use.


boxed:
Thanks!

PR #198, opened 2021-01-13
Simplify a few expressions

hukkinj1:


mutmut/__init__.py
```diff
@@ -1028,7 +1028,7 @@ def kill(process_):
                     if not line:
                         break
                     callback(line)
-        except (IOError, OSError):
+        except OSError:
```
hukkinj1, line 1031: See https://docs.python.org/3/library/exceptions.html#OSError

> Changed in version 3.3: EnvironmentError, IOError, WindowsError, socket.error, select.error and mmap.error have been merged into OSError, and the constructor may return a subclass.

mutmut/__init__.py
```diff
@@ -1090,7 +1090,7 @@ def write(self, s):
 
     modules_to_force_unload = {x.partition(os.sep)[0].replace('.py', '') for x in config.paths_to_mutate}
 
-    for module_name in list(sorted(set(sys.modules.keys()) - set(modules_before), reverse=True)):
+    for module_name in sorted(set(sys.modules.keys()) - set(modules_before), reverse=True):
```
hukkinj1, line 1093: `sorted()` returns a list, so no need to wrap in `list()`

mutmut/cache.py
```diff
@@ -158,7 +158,7 @@ def print_result_cache(show_diffs=False, dict_synonyms=None, print_only_filename
     print('')
 
     def print_stuff(title, mutant_query):
-        mutant_list = list(sorted(mutant_query, key=lambda x: x.line.sourcefile.filename))
+        mutant_list = sorted(mutant_query, key=lambda x: x.line.sourcefile.filename)
```
hukkinj1, line 161: sorted() returns a list, so no need to wrap in list()

mutmut/__init__.py
```diff
@@ -236,7 +236,7 @@ def number_mutation(value, **_):
 
 
 def string_mutation(value, **_):
-    prefix = value[:min([x for x in [value.find('"'), value.find("'")] if x != -1])]
+    prefix = value[:min(x for x in [value.find('"'), value.find("'")] if x != -1)]
```
hukkinj1, line 239: min() can take a generator

mutmut/__init__.py
boxed: Nice. I will try to remember that.

boxed:
Thanks!

Issue #201, opened 2021-03-26
--tests-dir and --paths-to-mutate have inconsistent string separation behavior


icncsx:
Can we include a help message for --tests-dir and --paths-to-mutate and highlight the fact that one is colon separated and the other is comma separated?

Happy to make a pull request :)

boxed:
I would propose that both should accept both comma and colon as separator. I will gladly take a PR!

icncsx:
Will do so. Thanks!

PR #206, opened 2021-04-11
Icncsx experimental pathstring

icncsx:
Hi, Anders. This PR is relevant to #201.

In brief, I am using regex to figure out if a path string is comma separated or colon separated. Depending on the match - either "," or ":", we are going to split the string on the relevant separation char.


boxed:
Clearly an improvement. Thanks!

PR #208, opened 2021-04-22
replace single with double back ticks

HerrMuellerluedenscheid:
The monospace highlighting did not work which makes browsing the README for the critical info cumbersome.


HerrMuellerluedenscheid:
The corrected is much easier to browser at least on github https://github.com/HerrMuellerluedenscheid/mutmut/tree/feature/formatting

Issue #193, opened 2020-12-11
[Feature Request] HTML output like pytest-cov


MartinThoma:
I would like to have a colorized output which shows all lines of code. The lines which contain surviving mutants should be colored read, the ones which contain killed mutants green (with the option to remove the highlighting via JavaScript).

I can generate an example with pytest-cov if that helps to communicate the idea what I'm looking for.

Would this be something you would also like?

boxed:
It's been suggested before. The problem I have with it is that it sounds good, and people would like it, but when they have it.. then what? Isn't it actually worthless? The only thing that actually hold value to kill mutants is to have the report we already have I think. 

Please tell me where my logic is flawed.

MartinThoma:
I'm currently running mutmut over a non-public project. It checks over 10k mutants. The structure of that project ... let's say, it could be improved. There are files that have A LOT of lines. The report generated by `mutmut html` lists many lines multiple times. The only thing I'm interested in per line is "did the test suite check this line properly?" - so if I see at least one failing mutant. In some cases, I want to see which one it is. But most of the time it's actually just "did I miss something here". It doesn't help me to see that line it multiple times. As there are so many things to be improved, I want to be able to prioritize the ones with the most impact. If I had just the whole code with highlighted lines, I could navigate the non-killed mutants more quickly and find issues more quickly.

Many mutants are showing not super interesting flaws of the test suite. But some are very important. By finding the important ones quickly I would also have an easier time to show my team the value of mutmut with a minimal pull request that addresses those.

MartinThoma:
By the way: When I asked the question, I didn't know that `mutmut html` existed. When I enter `mutmut --help` it doesn't show up:

```
$ mutmut --help
Usage: mutmut [OPTIONS] [COMMAND] [ARGUMENT] [ARGUMENT2]

  commands:

      run [mutation id]

          Runs mutmut. You probably want to start with just trying this. If
          you supply a mutation ID mutmut will check just this mutant.

      results

          Print the results.

      apply [mutation id]

          Apply a mutation on disk.

      show [mutation id]

          Show a mutation diff.

      show [path to file]

          Show all mutation diffs for this file.

      junitxml

          Show a mutation diff with junitxml format.

Options:
  --paths-to-mutate TEXT
  --paths-to-exclude TEXT
  --backup / --no-backup
  --runner TEXT
  --use-coverage
  --use-patch-file TEXT           Only mutate lines added/changed in the given
                                  patch file

[...]

PR #211, opened 2021-05-30
add coverage style html report

cleder:
Add a (quick and dirty) coverage style HTML report.
#193 
The generated HTML still needs some styling, but is usable.


boxed:
How does this improve on the existing report?

cleder:
image

cleder:
FYI: that is what I use for https://github.com/cleder/pygeoif/pull/19

PR #212, opened 2021-06-17
Add ability to update progress output only after test passes

PatrikValkovic:
Some CI (for example Travis) does not support "\r" sequence and the whole output is logged. That may result in too big file and fail of the CI, see my attempt to integrate mutmut: https://travis-ci.com/github/PatrikValkovic/grammpy/builds/229567121

I add `--no-progress` option to disable real-time output and progress is only shown after one mutation is tested.

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -805,7 +805,7 @@ class Config(object):
                  baseline_time_elapsed, test_time_multiplier, test_time_base,
                  backup, dict_synonyms, total, using_testmon, cache_only,
                  tests_dirs, hash_of_tests, pre_mutation, post_mutation,
-                 coverage_data, paths_to_mutate):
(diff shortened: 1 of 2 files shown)


boxed:
One print was just removed. What was the logic of that? Otherwise I like this change.

PatrikValkovic:
> One print was just removed. What was the logic of that? Otherwise I like this change.

If you refer to the line [1167](https://github.com/boxed/mutmut/pull/212/files#diff-ae59dfa5a700fd0fe571b27cd5990042000b93be2273bafbecf0a60989243ab9L1167) the call was removed as `progress.register` call the `print` method as the last statement. Unless there is some change to the state in the `update_mutant_status` (which I didn't find), the second `progress.print` is unnecessary.

PR #217, opened 2021-07-26
Add new argument to get all IDs of a certain result category

felixdivo:
This is useful if you want, for example, to re-run all surviving mutants after changing a lot: `for ID in $(mutmut result-ids survived); do mutmut run $ID; done`. Re-running all mutant checks was not an option (I takes days) and rerunning them manually would also be tedious since there were about 40 of them.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -447,3 +447,10 @@ def test_simple_output(filesystem):
     result = CliRunner().invoke(climain, ['run', '--paths-to-mutate=foo.py', "--simple-output"], catch_exceptions=False)
     print(repr(result.output))
     assert '14/14  KILLED 14  TIMEOUT 0  SUSPICIOUS 0  SURVIVED 0  SKIPPED 0' in repr(result.output)
+
+
+def test_simple_output(filesystem):
+    result = CliRunner().invoke(climain, ['run', '--paths-to-mutate=foo.py', "--simple-output"], catch_exceptions=False)
+    print(repr(result.output))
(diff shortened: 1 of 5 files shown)


tests/test_main.py
```diff
@@ -447,3 +447,10 @@ def test_simple_output(filesystem):
     result = CliRunner().invoke(climain, ['run', '--paths-to-mutate=foo.py', "--simple-output"], catch_exceptions=False)
     print(repr(result.output))
     assert '14/14  KILLED 14  TIMEOUT 0  SUSPICIOUS 0  SURVIVED 0  SKIPPED 0' in repr(result.output)
+
+
+def test_simple_output(filesystem):
+    result = CliRunner().invoke(climain, ['run', '--paths-to-mutate=foo.py', "--simple-output"], catch_exceptions=False)
+    print(repr(result.output))
+    assert CliRunner().invoke(climain, ['result-ids', "survived"], catch_exceptions=False).output.strip() == "1"
```
felixdivo, line 455: I didn't get the tests t run locally, but I tested the changes manually and they work. before merging, one would have to change this line to actually assert the correct behavior (14 IDs separated by spaces).

boxed:
I like it! What was the problem you had running the tests?

felixdivo:
Ah, I don't quite remember. Do you have time to look at it?

boxed:
@felixdivo (as evidenced by it taking 11 days to answer) not really. But I'll yolo merge this anyway.

PR #219, opened 2021-08-23
Followup to #217: Fix tests

felixdivo:
I felt bad for the state of #217, so I setup a new venv and the tests just seemed to work there. So I could fix the broken test case in this PR (along small other changes).


felixdivo:
CC @boxed

boxed:
Nice! Thanks!

PR #220, opened 2021-09-07
fix Windows process invocation when using venv

moltob:
`mutmut run` fails without error message if these conditions apply:

* Windows (10 in my case).
* Virtual environment used, where also pytest is installed.

The error message is not shown as the original Windows code caught `stderr` but then ignored the content of `process.stderr`. Instead of adding an explicit print for `stderr` I just redirected `stderr` to `stdout`, which is already printed.

Once I did that I saw the real issue:

```
Output:

C:\Program Files\Python39\python.exe: No module named pytest
```

So mutmut used my system Python interpreter instead of the one from the active venv. venv on Windows injects into `PATH`, see this line from the generated `activate.bat` for instance:

```
set PATH=%VIRTUAL_ENV%\Scripts;%PATH%
```

However from https://bugs.python.org/issue8557:

> [...] on Windows, PATH is only considered when shell=True is also passed.

After setting that, mutmut used my venv interpreter and worked as expected. I then also followed the [Popen recommendation](https://docs.python.org/3/library/subprocess.html#popen-constructor):

> If shell is True, it is recommended to pass args as a string rather than as a sequence.

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -987,9 +987,10 @@ def popen_streaming_output(cmd, callback, timeout=None):
     """
     if os.name == 'nt':  # pragma: no cover
         process = subprocess.Popen(
-            shlex.split(cmd),
+            cmd,
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE
+            stderr=subprocess.STDOUT,
+            shell=True,
         )
         stdout = process.stdout
     else:


boxed:
I don't see how this change fixes anything. It should be the same, except also have shell injection issues?

moltob:
Not sure how to respond to this without repeating the description. It fixes:

* Non-functioning output reporting of stderr on Windows.
* Non-functioning venv interpreter usage on Windows.

boxed:
The output redirect is fine.

I don't understand why `shlex.split(cmd)` on one side and `cmd` + `shell=True` on the other side fixes anything.

moltob:
Without it, the venv interpreter is not reachable, as `PATH` is not considered. The splitting was removed per Python doc recommandation.

boxed:
Why would `PATH` be ignored? It doesn't make sense. It should be exactly the same. And what doc are you talking about? Please provide a link.

moltob:
May I kindly ask whether you read the PR description? All reasoning and links are available there:

image

boxed:
Oh sorry. Ok, then I understand. Thanks for this PR!

disouzam:
I know this PR is old but, out of curiosity, what is required to make mutmut to work in Windows again (without WSL)?

I'm going through the WSL path but a path that was hard to go through for me. I'm not sure I have the capability to contribute with the changes but I'm curious because the error when I first tested last week was mostly related to resources package missing in Windows.

Before commenting here, I tried to find the history of removal of Windows from the support of mutmut and couldn't find yet.

boxed:
Mutmut 3 uses fork which is not available on windows. This makes mutmut radically faster.

Otto-AA:
I've created an issue where I've written down my thoughts on what would be needed to change: #397 

TL;DR: refactor the code to support different process start methods: `fork` for Linux+MacOS and `spawn` for Windows.

> This makes mutmut radically faster.

I could imagine that if processes are reused (so the pytest setup is only performed e.g. 10 times, regardless of how many mutants we have), that the performance on Windows could be quite close. I am not sure if reusing processes could have bad side effects that we cannot handle properly. So if someone is interested in working on this, I think it's worth a try to see how it works out.

PR #225, opened 2021-09-13
Add ``--enable-mutation-types`` and ``disable-mutation-types`` options

DudeNr33:
This PR adds two new command line options:

* ``--enable-mutation-types``: only the specified node types will be mutated
* ``--disable-mutation-types``: all node types except the ones specified will be mutated

Both options can take multiple types separated by comma, e.g.:

```console
mutmut run --disable-mutation-types=string,decorator
```

The types that can be specified are the keys of ``mutations_by_type``. If an unknown type is given to ``mutmut``, it will inform the user of the valid types that are available.

The options are exclusive and can not be combined.

Feedback is welcome!

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+2.3.0
+~~~~~
+
+* Add `--disable-mutation-types` and `--enable-mutation-types` to control what types of mutations are performed
+
 2.2.0
 ~~~~~
 
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -32,7 +32,7 @@ from time import time
 from parso import parse
 from parso.python.tree import Name, Number, Keyword
 
-__version__ = '2.2.0'
+__version__ = '2.3.0'
 
 
 try:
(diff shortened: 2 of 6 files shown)


HISTORY.rst
```diff
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+2.3.0
+~~~~~
+
+* Add `--disable-mutation-types` and `--enable-mutation-types` to control what types of mutations are performed
+
```
DudeNr33, line 8: I am just guessing that this would probably be part of a V2.3.0. 
If it is uncommon that the changelog and version is maintained by the submitter of the PR I can revert this change.

mutmut/__init__.py
```diff
@@ -32,7 +32,7 @@
 from parso import parse
 from parso.python.tree import Name, Number, Keyword
 
-__version__ = '2.2.0'
+__version__ = '2.3.0'
```
DudeNr33, line 35: Same comment as for ``HISTORY.rst``, I can revert this if necessary.

boxed:
Ah crap, I forgot I haven't moved this repo to github actions yet. I clicked the "ready" button because I thought it was the "run CI job for this new user" button.

HISTORY.rst
boxed: Any work I don't have to do it great :P Thanks!

boxed:
This looks fantastic!

boxed:
Thanks!

PR #226, opened 2021-09-15
Reenable evaluation of ``mutmut_config.init``

### LINKED ISSUE #224: ``mutmut_config.init()`` not called
I have troubles in getting my ``mutmut_config.py`` to do what I want it to.

The content ist:
```python
def init():
    from mutmut import mutations_by_type
    del mutations_by_type["string"]
    print(mutations_by_type)
```

When I run ``mutmut run``, the ``init()`` is not called.

What I could analyse so far:
* all the imports (from ``mutmut``'s ``__init__.py`` and ``__main__.py``) are done twice when running ``mutmut``, once when starting the program and once after the "2. Checking mutants" message
* At the first import, ``mutmut_config`` cannot be imported and is None. That is why the ``init()`` is not called. Only after the second import it can be loaded, and things like ``pre_mutation(context)`` work as expected

My directory layout is:
```
(.venv) tracematrix [try-mutmut●●] % tree
.
├── CHANGELOG.md
├── LICENSE
├── README.md
├── mutmut_config.py
├── poetry.lock
├── pyproject.toml
├── src
│   └── tracematrix
│       ├── item.py
│       ├── matrix.py
│       └── reporters
│           ├── __init__.py
│           ├── base_reporter.py
│           ├── csv_reporter.py
│           ├── html_reporter.py
│           └── template.html
└── tests
```

DudeNr33:
Since Python 3.4 the current working directory is not added to ``sys.path`` when running a console script.
Thus the ``init`` hook was no longer called when running ``mutmut``.

This PR fixes this and closes #224.

Also, I noticed that the test I added for the last PR was flaky, because the output contained the contents of a set.
I fixed it with this PR as well.

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -35,6 +35,8 @@ from parso.python.tree import Name, Number, Keyword
 __version__ = '2.3.0'
 
 
+if os.getcwd() not in sys.path:
+    sys.path.insert(0, os.getcwd())
 try:
     import mutmut_config
 except ImportError:
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -48,9 +48,6 @@ from mutmut.cache import print_result_cache, print_result_ids_cache, \
 from collections import namedtuple
 import re
 
-if os.getcwd() not in sys.path:
-    sys.path.insert(0, os.getcwd())
-
 
 def do_apply(mutation_pk, dict_synonyms, backup):
     """Apply a specified mutant to the source code
@@ -181,7 +178,7 @@ def main(command, argument, argument2, paths_to_mutate, disable_mutation_types,
         mutation_types_to_apply = set(mutations_by_type.keys())
         invalid_types = None
     if invalid_types:
-        raise click.BadArgumentUsage(f"The following are not valid mutation types: {', '.join(invalid_types)}. Valid mutation types are: {', '.join(mutations_by_type.keys())}")
+        raise click.BadArgumentUsage(f"The following are not valid mutation types: {', '.join(sorted(invalid_types))}. Valid mutation types are: {', '.join(mutations_by_type.keys())}")
 
     valid_commands = ['run', 'results', 'result-ids', 'apply', 'show', 'junitxml', 'html']
     if command not in valid_commands:
--- a/tests/test_mutmut_config_hooks.py
+++ b/tests/test_mutmut_config_hooks.py
@@ -1,0 +1,57 @@
+import os
+import sys
+import subprocess
+
+import pytest
+
+
+@pytest.fixture
+def basic_filesystem(tmpdir):
+    source_file = tmpdir / "foo.py"
+    source_file.write("def add(a, b): return a + b")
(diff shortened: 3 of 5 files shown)


mutmut/__init__.py
```diff
@@ -35,6 +35,8 @@
 __version__ = '2.3.0'
 
 
+if os.getcwd() not in sys.path:
+    sys.path.insert(0, os.getcwd())
```
DudeNr33, line 39: ``__init__`` is executed first even if ``__main__`` is the script that is actually passed to Python (or the entry point in this case), so we have to move this from ``__main__`` to here.

mutmut/__main__.py
```diff
@@ -181,7 +178,7 @@ def main(command, argument, argument2, paths_to_mutate, disable_mutation_types,
         mutation_types_to_apply = set(mutations_by_type.keys())
         invalid_types = None
     if invalid_types:
-        raise click.BadArgumentUsage(f"The following are not valid mutation types: {', '.join(invalid_types)}. Valid mutation types are: {', '.join(mutations_by_type.keys())}")
+        raise click.BadArgumentUsage(f"The following are not valid mutation types: {', '.join(sorted(invalid_types))}. Valid mutation types are: {', '.join(mutations_by_type.keys())}")
```
DudeNr33, line 181: Sorting the set creates deterministic output and therefore eliminates the flaky test.

tests/test_mutmut_config_hooks.py
```diff
@@ -0,0 +1,57 @@
...
+    
+    yield basic_filesystem
+
+    sys.path = original_path
+    os.chdir(original_dir)
+
+
+@pytest.mark.usefixtures("set_working_dir_and_path")
+def test_hooks(basic_filesystem):
+    subprocess.check_output(["python",  "-m", "mutmut", "run", "--paths-to-mutate=foo.py"])
```
DudeNr33, line 54: In order to realistically reproduce the error in the previous version of ``mutmut``, I have to change the working directory for the test. 
But this also means that if this test is run together with the other tests, the import of ``mutmut`` already happened when this test gets executed. That's why I used ``subprocess`` here, to create a new process where the ``__init__`` will be imported again.

tests/test_mutmut_config_hooks.py
```diff
@@ -0,0 +1,57 @@
...
+    assert (basic_filesystem / "init_hook").exists(), "init was not called."
+    assert (basic_filesystem / "pre_mutation_hook").exists(), "pre_mutation was not called."
+    assert (basic_filesystem / "pre_mutation_ast_hook").exists(), "pre_mutation_ast was not called."
```
DudeNr33, line 57: I thought adding some regression tests to verify the other hooks are correctly called as well won't hurt.

boxed:
Nice! Thanks again.

boxed:
I invited you to be a collaborator on mutmut. Feel free to decline if you want :P

DudeNr33:
Thanks! I accepted, mutation testing is really an interesting topic. 😄

PR #228, opened 2021-09-19
Use ``Click``'s subcommands

### LINKED ISSUE #56: Using click's subcommands
As discussed on #55, we should add support for `click`'s subcommand feature: http://click.palletsprojects.com/en/7.x/commands/. This provides greater control on arguments for each individual command, and should be a backwards compatible change. This would also improve extensibility of the CLI.

DudeNr33:
This PR refactors ``__main__.py`` to make use of ``Click``'s subcommand feature.

This has the following benefits:
* The end user can use the ``--help`` option more effectively, as he is only shown the options which are actually necessary for this command
* The arguments can now be named according to their intent. As they were used for different subcommands previously, they were generically named ``argument`` and ``argument2``.
* The ``main()`` function (now named ``do_run()``) is now only responsible for handling the logic of the ``run`` command. Before it also had additional responsibilities like checking and evaluating the input parameters from the command line, and delegating/executing the logic for the other commands.

This change is _almost_ backwards compatible. The only thing I could not get to work the same way as before is the ``mutmut --version`` command. This is now a normal command and has to be invoked with ``mutmut version``.
It may be discussed if this should lead to a bump in the major version. I left the version number as it is right now.

I added test cases for the commands and options that were not covered so far (with exception for the ``--dict-synonyms`` option). They are in the same integration test style as the existing tests. 

While assessing the existing commands and options, I noticed a few things in the code that might be addressed in new issues. I left them untouched for now:
* The previous ``main`` also deals with command ``diff`` as synonym for ``show``, which cannot be passed through the CLI because is was not in the ``valid_commands`` list. I did not include a subcommand for ``diff`` as it could not be executed before as well.
* ``Config`` stores ``backup`` as instance attribute, which is never accessed. It is only used by the ``apply`` command, but here it is used directly and is not read from the ``Config`` instance. Should this be removed?
* The ``--cache_only`` option does not seem to do anything.
* ``mutmut show all True`` is not working as intended - it only shows the number of surviving mutants, not the filenames.
* Some of the tests that rely on the ``filesystem`` fixture are somehow interdependent - ``test_html_output`` failed if using the existing ``filesystem`` fixture and it is run together with other tests because all mutants are skipped/untested, but if run standalone this does not happen.
* In ``run``: it is probably not necessary to guess the paths to mutate if an argument is given. 

Feedback is welcome. 

Closes #56

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,10 @@ Changelog
 
 * Fixed error where ``mutmut_config.init()`` was not called when running without explicitly having set ``PYTHONPATH``
 
+* Use ``Click``'s subcommand feature to refactor the command line interface. For the end user, this can now run ``mutmut [COMMAND] -h``
+  to check which parameters are relevant to this specific subcommand. The change is backwards compatible, and all existing commands
+  work the same as before, with the exception of ``mutmut --version``, which now has to be ``mutmut version``.
+
 2.2.0
 ~~~~~
 
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -808,7 +808,7 @@ def run_mutation(context: Context, callback) -> str:
 class Config(object):
     def __init__(self, swallow_output, test_command, covered_lines_by_filename,
                  baseline_time_elapsed, test_time_multiplier, test_time_base,
-                 backup, dict_synonyms, total, using_testmon, cache_only,
+                 dict_synonyms, total, using_testmon, cache_only,
                  tests_dirs, hash_of_tests, pre_mutation, post_mutation,
                  coverage_data, paths_to_mutate, mutation_types_to_apply, no_progress):
         self.swallow_output = swallow_output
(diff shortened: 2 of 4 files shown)


boxed:
> Config stores backup as instance attribute, which is never accessed. It is only used by the apply command, but here it is used directly and is not read from the Config instance. Should this be removed?

Good catch. Yes.

> The --cache_only option does not seem to do anything

I think it's supposed to just update the cache, ie the database of where line numbers are and such. Mutmut can handle if you add a line in the beginning of the file and all line numbers move one step. That's why this command exists. It might be a bit subtle what it actually does :P

Great changes! You can go ahead and merge when you're ready.

DudeNr33:
Ok, I removed the ``backup`` variable from ``Config``, which also means that it is not needed any longer as an option for the ``run`` command as well.

As for ``--cache_only``: I double checked the code and cannot find where it is used.
It is passed as an option to the ``run`` command, gets handed down to the ``do_run`` (former ``main``) method, and there passed to the ``Config`` object. But it is never accessed later on.
I'll leave it in there, maybe this had an effect in an earlier version and was mistakenly deleted and this is a regression? 

But for the scope of this PR I'm done and will merge now. I'll open new issues for the things I noticed so they won't be forgotten.

boxed:
🎉

PR #231, opened 2021-09-20
Add GitHub actions

DudeNr33:
This adds a CI pipeline to (at least partially) replace the Travis pipeline.
What is not included yet is the deploy stage. I'd personally use a separate workflow for this, so this could be done in a future PR.
I also included a linting stage (and noticed I had quite a few "style errors" in the last PRs, should've run the ``tox -e lint`` earlier 😁).

Let's see if this action already triggers through this PR to see if it works as intended.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,25 @@
+name: CI
+on:
+  pull_request:
+    branches:
+    - master
+  push:
+    branches:
+    - master
+
+jobs:
+  ci:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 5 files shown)


boxed:
The deploy thing can be dropped I think. 

I personally abhor linting that fails the build. It's just annoying. A lint stage on PRs that is "FYI" instead of a hard fail is good though. 

Thanks for doing this! You're on a roll!

DudeNr33:
I'll try to do this with the ``continue-on-error`` option, but honestly I don't know yet if this will work as intended (see also https://github.com/actions/toolkit/issues/399 for some discussion about this).
If it does not work out, we can remove this stage again.

I also fear that we have to merge this PR to check if the pipeline succeeds.

boxed:
Yea. My experience with github actions is that you push a hundred times to get stuff to work :) Just accepy it and go go

PR #232, opened 2021-09-21
Remove ``--cache-only`` option and second argument to ``mutmut show``

### LINKED ISSUE #229: ``mutmut show all <filename>`` and ``mutmut show <filename>`` equivalent?
During the last PR I was a bit confused what the second argument to ``mutmut show <argument> <argument2>`` should be.

The code is the following:
```python
def print_stuff(title, mutant_query):
    mutant_list = sorted(mutant_query, key=lambda x: x.line.sourcefile.filename)
    if mutant_list:
        print('')
        print("{} ({})".format(title, len(mutant_list)))
        for filename, mutants in groupby(mutant_list, key=lambda x: x.line.sourcefile.filename):
            if print_only_filename is not None and print_only_filename != filename:
                continue

            if only_this_file and filename != only_this_file:
                continue
```

At first I thought that the second argument means "print only the filenames that have surviving mutants".
But reading the commit message for the first ``if`` in the ``for`` loop, I guess that the intention was to display all mutant diffs for this specific file:
"Display all diffs for surviving mutants for a specific file with `mutmut show all path/to/file.py`"

However, this is also possible by just running ``mutmut show path/to/file.py``. 
Hence the two ``if`` blocks effectively do the same thing. 

The question is: how should we proceed? 
Just drop the second argument, or use it to actually just list the names of the affected files?

### LINKED ISSUE #230: ``--cache-only`` option does not do anything?
``--cache-only`` can be passed as an option to the ``run`` command.
However, this option seems to have no effect. It is passed into the ``Config`` class, but never accessed again. 

From the discussion in PR #228:
> I think it's supposed to just update the cache, ie the database of where line numbers are and such. Mutmut can handle if you add a line in the beginning of the file and all line numbers move one step. That's why this command exists. It might be a bit subtle what it actually does :P

DudeNr33:
Closes #229 
Closes #230

This is also a great chance to see if the new CI workflow is working as intended.
I inadvertently broke some tests with the last PR because ``flake8`` was not happy with some trailing whitespace in the expected output.
I reverted those changes, but this should now lead to a failure in the linting stage. Let's see if the ``continue-on-error`` option works as intended!

But for the final solution I will try to find a way to satisfy ``flake8`` without breaking the tests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -8,18 +8,6 @@ on:
     - master
 
 jobs:
-  lint:
-    runs-on: ubuntu-latest
-    continue-on-error: true
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
-        with:
-          python-version: 3.9
-      - name: Setup tox
-        run: pip install tox
-      - name: Run linting
-        run: tox -e lint
   test:
     strategy:
       fail-fast: false
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -808,7 +808,7 @@ def run_mutation(context: Context, callback) -> str:
 class Config(object):
     def __init__(self, swallow_output, test_command, covered_lines_by_filename,
                  baseline_time_elapsed, test_time_multiplier, test_time_base,
-                 dict_synonyms, total, using_testmon, cache_only,
+                 dict_synonyms, total, using_testmon,
                  tests_dirs, hash_of_tests, pre_mutation, post_mutation,
                  coverage_data, paths_to_mutate, mutation_types_to_apply, no_progress):
         self.swallow_output = swallow_output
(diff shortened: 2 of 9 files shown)


DudeNr33:
Ok, so it seems we have three options for the lint stage:

1. Keep it as it is, but from what I've understood from the discussion on the last PR that is not desired as it will show a big red error like in this example here
2. Change the command ``tox -e lint`` to ``tox -e lint || true``. This way the workflow will always show green for the lint stage. I personally do not like this, as this only hides the errors and gives a false sense of "everything is fine". In this case I'd rather have option 3:
3. Remove the lint stage completely

Unfortunately Github Actions only know success and failure. A "warning" or "unstable" result would be nice, but it is what it is. 😅

boxed:
I say kill the lint stage then. Hmm. Or maybe the github action could just write a comment in the PR? That's be great.

PR #233, opened 2021-09-26
Improve test selection through ``pre_mutation(context)``

### LINKED ISSUE #227: Add option to take coverage contexts into account
``coverage.py`` is able to capture [dynamic contexts](https://coverage.readthedocs.io/en/coverage-5.5/contexts.html#dynamic-contexts), which means it can track which test exercised which parts of the code under test.

As ``mutmut`` is already able to read the coverage database, it could significantly improve the execution times if it would take this context data into account and only run the tests that actually cover the mutated line.

If no context was recorded or the context is not the name of a test function (e.g. a user-defined static context for example), ``mutmut`` should run the whole test suite as a fallback. 
This feature should probably also be optional.

DudeNr33:
Closes #227

After looking at the code and the last example in the discussion of the issue, I realised that most of the necessary logic was already present.
The changes to the code are minimal:

* Read the coverage contexts from the ``.coverage`` file instead only the covered lines
* Reset the ``config.test_command`` after each mutation run so that the user does not have to take care of this
* Add an option which tells ``mutmut`` to re-run the whole test suite if the mutant survives after running just the subset of tests.

The rest is documentation to give some concrete examples how to do this, not only for selection based on coverage contexts, but also other cases.

I decided to reuse the existing ``pre_mutation(context)`` hook, as a new function would just lead to duplications in the code and ``pre_mutation`` is reasonably suited for this. The current docs already describes this hook as 

> pre_mutation gets called before each mutant is applied and tested.

Let me know what you think!

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -12,6 +12,9 @@ Changelog
   to check which parameters are relevant to this specific subcommand. The change is backwards compatible, and all existing commands
   work the same as before, with the exception of ``mutmut --version``, which now has to be ``mutmut version``.
 
+* You can now set the ``context.config.test_command`` in the ``mutmut_config.pre_mutation(context)`` hook to select the relevant subset of tests.
+  Coverage contexts are now accessible in ``config.coverage_data`` to help with the selection.
+
 2.2.0
 ~~~~~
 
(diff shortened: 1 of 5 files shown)


boxed:
> Reset the config.test_command after each mutation run so that the user does not have to take care of this

Damn. That was a pretty big bug imo! Nice of you to spot it!

boxed:
Wow. I'm pretty impressed by the docs you wrote. This is awesome stuff!

PR #235, opened 2021-10-05
Fix inconsistent output of ``mutmut show`` after using ``--enable-mutation-types``

### LINKED ISSUE #234: Bug with --disable-mutation-types and --enable-mutation-types
Hi I have been testing out the new commands since they are quite useful. However I found some bugs related to them. 

Here is a testProgram for mutation purpose 

```
def testfunction(): 
    array['a'] = 123
```

I run this command after building v2.3 `mutmut run --enable-mutation-types=number --paths-to-mutate testFunction.py` 
followed by 
`mutmut show all`

and here is the output:

```
To apply a mutant on disk:
    mutmut apply <id>

To show a mutant:
    mutmut show <id>


Survived � (1)

---- testFunction.py (1) ----

# mutant 1
--- testFunction.py
+++ testFunction.py
@@ -1,3 +1,3 @@

 def testfunction():
-    array['a'] = 123
+    array['XXaXX'] = 123
```

This seems like a string mutation and not a number mutation, the correct number mutation after taking a brief look at the source code shoule be something like `array['a'] = 124`

You can get a similar "wrong" output when you run `--disable-mutation-types=string`

I am not very familar with the code base but I think this bug has something to do with overwriting mutations? see these lines https://github.com/boxed/mutmut/blob/cc7e7825a8a2ca0d5abb5f4748aa8c8ccea7f0f7/mutmut/__init__.py#L629-L643

DudeNr33:
Closes #234

``mutmut show <id>`` did not show the same mutant that was created and tested with ``mutmut run`` because the ``index`` of the mutation did not match. 
By moving the logic if a mutation should be performed into ``Context.should_mutate(node)``, this is fixed.

Note: in the issue we discussed that it might be a good idea to mark the mutants that are not applied and run due to ``--enable-mutation-types`` or ``--disable-mutation-types`` as ``SKIPPED``.
This is not part of this PR, as implementing it would be tricky. Identifying the mutants and actually running them and setting a status is decoupled, and we'd have to add some means of tagging a mutant as "not-to-be-run" in some way.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -101,6 +101,14 @@ def test_nothing(): assert True
 
     create_filesystem(tmpdir, foo_py, test_py)
 
+    yield tmpdir
+
+    # This is a hack to get pony to forget about the old db file
+    # otherwise Pony thinks we've already created the tables
+    import mutmut.cache
+    mutmut.cache.db.provider = None
+    mutmut.cache.db.schema = None
+
 
 def create_filesystem(tmpdir, file_to_mutate_contents, test_file_contents):
     test_dir = str(tmpdir)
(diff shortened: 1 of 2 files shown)


tests/test_main.py
```diff
@@ -601,6 +609,48 @@ def test_enable_and_disable_mutation_type_are_exclusive():
...
+ def foo(a, b):
+-    result = a + b
++    {expected_mutation}
+     return result
+""" in result.output
+
+
+def test_run_multiple_times_with_different_mutation_types(filesystem):
+    """Running multiple times with different mutation types enabled should append the new mutants to the cache without
+    altering existing mutants."""
```
DudeNr33, line 634: I wondered if running different mutation types one after the other works without breaking the cache. Luckily it does!

PR #239, opened 2021-11-03
Change very small or very large numbers by a relative amount

### LINKED ISSUE #238: Generation of unary plus in exponent is un-killable
Hello, thanks for the lovely tool!

mutmut generates mutants like the following, which are identical to the original code and thus un-killable (immortal?):

```diff
--- mvce/mvce.py
+++ mvce/mvce.py
@@ -1,3 +1,3 @@
 def multiply_by_1e100(number):
-    return number * 1e100
+    return number * 1e+100
```

Here's a test so that you can reproduce the whole thing:

```py
from mvce import multiply_by_1e100


def test_multiply_by_1e100():
    assert multiply_by_1e100(4) == 4e100
```

`1e16` seems to be the smallest number to which this happens. Smaller numbers get expanded from e.g. `1e2` to `101.0`

ZedThree:
This avoids (some) problems with unkillable mutants, for example:

    1e16 -> 1e+16

which now becomes

    1e16 -> 2e+16

Fixes #238


boxed:
Thanks!

PR #242, opened 2022-01-24
add pyproject support

### LINKED ISSUE #169: Should support pyproject.toml as well as setup.cfg
My new projects don't have a setup.cfg, only a pyproject.toml and tox.ini.
Could mutmut be modified to read config options from either file?

sed-i:
This PR adds support for reading the config from `pyproject.toml` ([PEP 621](https://www.python.org/dev/peps/pep-0621/)).
Precedence is given to `pyproject.toml` over `setup.cfg`, which is common practice, for example such as in [Mypy](https://mypy.readthedocs.io/en/stable/config_file.html#the-mypy-configuration-file).

Existing users of `setup.cfg` will most likely not be affected if they also have `pyporject.toml`: the code in this PR gives precedence to `pyproject.toml` only if a `tool.mumut` section is found in the `pyproject.toml` file.

Fixes #169

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -7,11 +7,8 @@ import re
 import shlex
 import subprocess
 import sys
-from configparser import (
-    ConfigParser,
-    NoOptionError,
-    NoSectionError,
-)
+import toml
+from configparser import ConfigParser
 from copy import copy as copy_obj
 from functools import wraps
 from io import (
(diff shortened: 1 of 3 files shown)


boxed:
Thanks!

MartinThoma:
It might be worth to update the docs https://mutmut.readthedocs.io/en/latest/#install-and-run (search for "setup.cfg")

PR #255, opened 2022-08-15
Fix formatting

Zac-HD:
Trivial fix, [see here for the current rendering](https://mutmut.readthedocs.io/en/latest/#selection-based-on-coverage-contexts).


boxed:
Thanks

Zac-HD:
Hmm, looks like the readthedocs build is failing with a Python 2 error: https://readthedocs.org/projects/mutmut/builds/

boxed:
Thanks for the headsup. I just needed to set rtd to use python3 and it went through fine.

Issue #260, opened 2022-10-16
Line numbers in junitxml report seem to be off by one


srazzell:
<img width="649" alt="Screen Shot 2022-10-16 at 10 43 49 AM" src="https://user-images.githubusercontent.com/84470707/196050075-c40b9427-2cca-4852-af11-d5f099320db9.png">

The "line" attribute of the testcases in the junitxml report seem to be off by one. 

I ran `mutmut run --paths-to-mutate "mutations/__main__.py"`, then ran `mutmut junitxml`, and I get this output:
`<testcase name="Mutant #12" file="mutations/__main__.py" line="21">`
[bug.txt](https://github.com/boxed/mutmut/files/9795325/bug.txt)

I attached a screenshot to show the line number for the file mutmut is running on. This happens with every mutation in the xml report.

boxed:
Hmm. I guess just a `+1` on https://github.com/boxed/mutmut/blob/master/mutmut/cache.py#L237

Could you try that and see if that fixes it?

srazzell:
Yes, that does fix it. I'm wondering where that number is calculated. Is it here?

https://github.com/boxed/mutmut/blob/master/mutmut/cache.py#L368

Line numbers should start at 1 right? So maybe adding `+1` on line 368, or starting `i` at `1` instead of 0 in the enumerate?

boxed:
Hmm. Yea try that.

boxed:
That would be great!

srazzell:
PR:
https://github.com/boxed/mutmut/pull/261

srazzell:
PR was merged

PR #261, opened 2022-10-22
Fixing junitxml line number, daemon warning, and missing argument in unit test

srazzell:
Issue: https://github.com/boxed/mutmut/issues/260

Changes:
1. Add to one to junitxml report line numbers to fix above issue
2. Fix a unit test that was missing the no_progress argument
3. Fix a daemon warning from the unit tests
4. Separate creating the junitxml report from printing it, so that it can be tested easier in the future.

All unit tests pass with these changes.


boxed:
Thanks!

PR #262, opened 2022-10-22
Fix bug with init db on wrong function

srazzell:
If you try to run it directly, the new function `create_junitxml_report` throws an error since the db is not initalized.

I'm sorry I didn't find this before my previous pull request.


boxed:
Thanks. And no problem :P

PR #264, opened 2022-10-26
Changing mutation decode from UTF-8 to Latin1

sidneydemoraes:
This change allows MUTMUT to analyze code which contains log messages and comments written in Latin-based languages without breaking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,7 @@ docs/tri*.rst
 
 # tests
 results/
+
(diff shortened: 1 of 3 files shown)


sidneydemoraes:
Added more  files than I should. Sorry for that. 
I'll send a new one another time.

boxed:
That change would break the vast majority of uses. Python source files should be utf8!

Issue #183, opened 2020-09-13
Mutmut stops running 


saikatG:
Hi,

Mutmut stops running after a while and outputs the following error:

```
Exception in thread queue_mutants:                                                                                                                                                                                 
Traceback (most recent call last):
  File ".../python3.7/site-packages/mutmut/__init__.py", line 712, in queue_mutants                                                                                             
    mutants_queue.put(('mutant', context))                                                                                                                                                                         
  File "/.../python3.7/multiprocessing/queues.py", line 81, in put                                                                                                               
    assert not self._closed, "Queue {0!r} has been closed".format(self)                                                                                                                                            
AssertionError: Queue <multiprocessing.queues.Queue object at 0x7fe5a0b69f50> has been closed                                                                                                                      
                                                                                                                                                                                                                   
During handling of the above exception, another exception occurred:                                                                                                                                                
                                                                                                                                                                                                                   
Traceback (most recent call last):                                                                                                                                                                                 
  File "/.../python3.7/threading.py", line 926, in _bootstrap_inner                                                                                                              
    self.run()                                                                                                                                                                                                     
  File "/.../python3.7/threading.py", line 870, in run                                                                                                                           
    self._target(*self._args, **self._kwargs)                                                                                                                                                                      
  File "/.../python3.7/site-packages/mutmut/__init__.py", line 715, in queue_mutants                                                                                             
    mutants_queue.put(('end', None))                                                                                                                                                                               
  File "/.../python3.7/multiprocessing/queues.py", line 81, in put                                                                                                               
    assert not self._closed, "Queue {0!r} has been closed".format(self)                                                                                                                                            
AssertionError: Queue <multiprocessing.queues.Queue object at 0x7fe5a0b69f50> has been closed                                                                                                                      
                                                                                                                                                                                                                   
Traceback (most recent call last):                                                                                                                                                                                 
  File "/.../python3.7/multiprocessing/queues.py", line 242, in _feed                                                                                                            
    send_bytes(obj)                                                                                                                                                                                                
  File "/.../python3.7/multiprocessing/connection.py", line 200, in send_bytes                                                                                                   
    self._send_bytes(m[offset:offset + size])                                                                                                                                                                      
  File "/.../python3.7/multiprocessing/connection.py", line 398, in _send_bytes                                                                                                  
    self._send(buf)                                                                                                                                                                                                
  File "/.../python3.7/multiprocessing/connection.py", line 368, in _send                                                                                                        
[...]

PR #269, opened 2022-12-11
#183 Mutmut stops running

yanekk:
Hi! I stumbled upon this problem of hanging mutmut on process cycling. I'm experimenting with mutmut as a part of a CI pipeline After some debugging, I discovered that the actual hanging behaviour was caused by hanging `queue_mutants` thread, which hangs at `put` method of the mutant queue because it was full and there was no consumer to get the commands. 

However, the root cause (in my case, at least) was the race condition between results consumer loop (in `run_mutation_tests`) and mutant worker loop (in `check_mutants`). 

Results consumer loop depends its execution on `is_alive` method of the current worker. If worker sends `cycle` command and terminates before consumer loop finishes updating status of previous mutant (which happened in my case), the loop ends without attempting to recreating the worker, hence finishing execution of `run_mutation_tests` function. Such behavior caused dangling threads to hang the execution of parent process, hence making whole process both unresposive and unfinished.

The fix I propose is to remove dependency on the `is_alive` method of current worker process, by changing it to infinite loop. By introducing this change, the only way to end the loop is to receive `end` command, but with the `finally` sending it in worker loop, such possibility is quite unlikely.

Additionally, I added a test which exhibits this behaviour, however it's quite mock-heavy to make it focused on a race condition in question instead of making it full end-to-end test.

Tested on:
 - Windows 10, x86, Python 3.9.13
 - Ubuntu (WSL), x86, Python 3.8.10

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -1123,6 +1123,7 @@ def hammett_tests_pass(config, callback):
 
     return returncode == 0
 
+CYCLE_PROCESS_AFTER = 100
 
 def run_mutation_tests(config, progress, mutations_by_file):
     """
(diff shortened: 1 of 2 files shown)


boxed:
Thanks a lot for this!

PR #271, opened 2023-01-28
Fix issue with `TypeError` for empty file diffs

yanekk:
Hi!

Here's the fix for the problem I stumbled upon when I was doing some experiments on running mutation tests based on git patch. The problem was `whatthepatch` library returns `None` instead of empty list as a value for `diff.changes` if file has no changes (new empty file, renamed empty file, removed empty file - tested on MacOS and Windows 10), which made mutmut to crash because of the set comprehension requirement in `__init__.py`. I tested it on current version of whatthepatch (1.0.3) and one from requirements (0.0.6), but for both versions the behaviour was the same. 

Unfortunately, not only it didn't parse the patch, but also crash `mutmut`. Therefore, I thought it would be good idea to fix it, since without it it would require for my use case to generate list of changed lines externally as a coverage report and pass it with the `--use-coverage` parameter. 

As before, I added a regression unit test suite to make sure it won't happen again.

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -1220,7 +1220,7 @@ def read_patch_data(patch_file_path):
 
     return {
         diff.header.new_path: {change.new for change in diff.changes if change.old is None}
-        for diff in diffs
+        for diff in diffs if diff.changes
     }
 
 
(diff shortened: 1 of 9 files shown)


boxed:
Thanks!

PR #273, opened 2023-02-09
Introduce mutation of f-strings

yanekk:
Hi, I noticed there's only partial support for mutation of f-strings. For example, given following f-string:
```python
f'this is {True} an f-string'
``` 
we got only one mutation:
```python
f'this is {False} an f-string'
```
so the only part that's mutated is the `True` in curly brackets (I know it doesn't make sense, but it was easiest case to exemplify). The f-string doesn't get the same `XX`-type mutation as plain string does, which might find few new issues if f-string is used extensively for string formatting in the project. This PR is an attempt to introduce f-string mutation to mutmut. Once introduced, example above gives two mutations instead of one:
```python
f'this is {False} an f-string'
f'XXthis is {True} an f-stringXX'
```
So we mutate not only expressions within the string, but also the string itself. Unit tests passing locally, I also tested it manually and all found issues are fixed.

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -27,7 +27,7 @@ from threading import (
 from time import time
 
 from parso import parse
-from parso.python.tree import Name, Number, Keyword
+from parso.python.tree import Name, Number, Keyword, FStringStart, FStringEnd
 
 __version__ = '2.4.3'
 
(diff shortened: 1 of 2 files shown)


boxed:
Nice!

PR #277, opened 2023-03-27
Allow newer versions of junit-xml

TalAmuyal:
`junit-xml` version 1.9 provides a Wheel which is faster to install.
This is relevant for `mutmut` as well as downstream users of `mutmut`.

Please consider merging it so I could bump `junit-xml`'s version in my projects.

Thanks :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,5 @@ glob2
 parso
 click
 pony
-junit-xml==1.8
+junit-xml>=1.8,<2
 toml


boxed:
You don't need to install mutmut in the same venv as your project.

TalAmuyal:
You are correct, I'm moving it out of the main lockfile.
Thanks for merging this anyway, I appreciate that :)

PR #288, opened 2023-09-12
Use dataclasses for Config & RelativeMutationID

EJahren:
I am adding type hints while trying to understand the code, and dataclasses made that a bit easier.

 The only changed behavior is that `repr(config)` is no longer `<__init__.Config object at ID>` and `hash(config)` is no longer `id(config)`.

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -800,7 +800,9 @@ def run_mutation(context: Context, callback) -> str:
             return BAD_TIMEOUT
 
         time_elapsed = time() - start
-        if not survived and time_elapsed > config.test_time_base + (config.baseline_time_elapsed * config.test_time_multipler):
+        if not survived and time_elapsed > config.test_time_base + (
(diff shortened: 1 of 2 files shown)


mutmut/__init__.py
```diff
@@ -802,10 +792,6 @@ def run_mutation(context: Context, callback) -> str:
         start = time()
         try:
             survived = tests_pass(config=config, callback=callback)
-            if survived and config.test_command != config._default_test_command and config.rerun_all:
-                # rerun the whole test suite to be sure the mutant can not be killed by other tests
-                config.test_command = config._default_test_command
-                survived = tests_pass(config=config, callback=callback)
```
boxed, line 808: why is this removed?

EJahren:
Hi @boxed! Sorry about the noise, I completely misunderstood how `_default_test_command` works, but I believe the current changes should be uncontroversial.

boxed:
👍

PR #292, opened 2023-10-13
Raise FileNotFoundError if no test dirs are found

khink:
Before this change, a `TypeError: 'NoneType' object is not iterable` would occur. The new exception is more helpful.


boxed:
Thanks!

PR #299, opened 2024-01-05
Update to_xml_string to to_xml_report_string to fix deprecation warning

jimrybarski:
`junit_xml` has renamed `TestSuite.to_xml_string()` to a function `to_xml_report_string()`. Currently, running pytest will trigger a `DeprecationWarning`. This commit simply updates the usage to the new function.


boxed:
Thanks!

PR #300, opened 2024-01-08
document configuration with pyproject.toml

brodybits:
This could have helped me as I have been working on a project with no `setup.cfg` file and am still pretty new to the whole Python build & release tooling. Thanks!


boxed:
Thanks!

PR #310, opened 2024-03-15
bugfix: Untested/skipped in show all does not show skipped

Century-ss:
@boxed 

**Bug**: When I run the `mutmut show all` command, the "Untested/skipped" display does not include skipped.
I have fixed it so that skipped is also included.

I would appreciate it if you could check.


boxed:
Thanks!

PR #313, opened 2024-04-04
Normalize paths from patch file

### LINKED ISSUE #311: Use patch file on Windows never executes mutants
Using the option `--use-patch-file` on Windows is not possible.
Patch file and the resulting `Config.covered_lines_by_filename` array do use unix paths separators "/". While the `Context.filename` uses windows paths separators "\\".

This causes `should_exclude` always to return True, and therefore no mutants are tested at all.

```py
Config.covered_lines_by_filename = {
  'src/my-pkg/foo.py': { 1, 2, 42 }
}
Context.filename = 'src\\my-pkg\\foo.py'
```

FlTr:
os.path.join will use either "/" or "\" as path separator, depending on running host system. Patch files use UNIX separators.

Fixes #311

### DIFF
--- a/mutmut/__init__.py
+++ b/mutmut/__init__.py
@@ -1220,7 +1220,7 @@ def read_patch_data(patch_file_path: str):
         diffs = whatthepatch.parse_patch(f.read())
 
     return {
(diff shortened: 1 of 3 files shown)


boxed:
Thanks!

PR #314, opened 2024-04-04
Add option to html command to specify output directory

FlTr:
Default directory stays "html", added "--directory" option similar to coverage package.


boxed:
There was another merged PR in between that added skipped mutants to the html output.

FlTr:
Thanks for the hint, just rebased it.

PR #318, opened 2024-04-22
Add note about mutmut_config.py cache issue

butvinm:
After spending some time troubleshooting why my new whitelist rules are not applied, I discovered that changes to mutmut_config.py don't invalidate the cache, and mutmut just returns the previous result without taking the new rules into account. Therefore, I suggest adding a note to the documentation to clarify this behavior.


boxed:
Thanks!

PR #320, opened 2024-05-21
Iterator pattern

InteVleminckx:


boxed:
I hope this was a school assignment to teach you guys that using tools that blindly look at cyclomatic complexity won't lead to anything good :P

PR #324, opened 2024-06-13
Run CI on 3.10 and 3.11

sobolevn:
Let's see what the CI will say about it :)


sobolevn:
3.12 failed. @boxed, I propose merging this without 3.12 for now.
Later it can be added when 3.12-specific bug is fixed.

Issue #302, opened 2024-01-24
Issue with mutmut mutation in presence of type hints


pablo-snz:
Hi! I'm wondering if there's a direct way to exclude type hints within mutmut. For example, in the following code, I'm getting errors because mutmut tries to mutate the "|" operator to its counterpart "&":

```{python}
@dataclass
class Specification:
    filter: Criteria | None = None  # Mutates to => Criteria & None = None
    order: Order | None = None  # Also happens in all of the cases below
    limit: int | None = None
    offset: int | None = None
```
Plus it becomes more complex in a scenario like this where I can't just whitelist the entire line:

```{python}
user: User | None = self.repo.match(Specification(filter=Equals("email", user.email) | Equals("username", user.username)))
```

On the other hand, if mutmut cannot handle this right now in a straightforward manner, would it be possible for me to contribute to the repository by adding this functionality?

boxed:
Hmm. I guess in general it would be a good idea to ignore type hints totally, except they are used for behavior in stuff like pydantic. 

I am very much open to a PR. Not sure what should be done though. The smallest change would be for the operator mutation for or to not trigger if inside an annotation. A big ugly but would fit this case.

efine-vivodyne:
I just started playing with mutmut today, and this was one of the immediate pain points I ran into.  I would also very much be interested in better support for type hints.

I see the potential issue with pydantic, but also I've never run across any code actually using `&` in a type hint.  I think a flag in the configuration that said "never attempt to mutate a `|` to a `&` in a type hint" with be something I would gladly turn on and never have a second thought that I'm risking less thorough mutation testing of my code

efine-vivodyne:
I don't know if there's another issue for this already, but another odd thing I saw was

```
-    port: int | None = None
+    port: int | None = ""
```

I understand in general that mutating default values is a good thing to probe.  But if someone is using type hints in their code, presumably they're also using a type-checker, which would flag an attempt like this to assign a default value that doesn't match the type hint.  This seems like a false positive that mutmut is creating

EdgyEdgemond:
Have taken advantage of mutmut_config.py to allow skipping lines with `# nomut` comment, would this be a simple work around (default premutation)?

```
from mutmut import Context


def pre_mutation(context: Context) -> None:
    line = context.current_source_line.strip()
    if line.contains("# nomut"):
        context.skip = True
```

[...]

PR #327, opened 2024-06-28
Skip union mutations for type hints

EdgyEdgemond:


boxed:
This is not the correct approach, as i said in #302

PR #333, opened 2024-08-26
Make the bash script easily copy-pasteable

vivganes:
Right now, the bash script to run mutations is not a code block.  There is a potential of people copy-pasting wrong thing.

When we show it as a code block, then copy-pasting is accurate and also easy.


boxed:
Thanks

Issue #349, opened 2024-11-21
Exit 1 with "FAILED" and no explanation after forced fail test


WillGibson:
I thought it might be because I have skipped a few tests for now because they don't play ball when running with mutmut, but I commented them out and the result was the same.

Should [this line](https://github.com/boxed/mutmut/blob/main/mutmut/__main__.py#L892) be...

```
if runner.run_forced_fail() != 0
```

It's not obvious to me what the purpose or expected behaviour of the forced failed run is and it [doesn't seem to be mentioned in the documentation](https://mutmut.readthedocs.io/en/latest/search.html?q=forced).

[I don't think our code resets os.environ anywhere](https://github.com/search?q=repo%3Auktrade%2Fplatform-tools%20os.environ&type=code).

## Steps to replicate

```
# Setup
git clone git@github.com:uktrade/platform-tools.git
git checkout ad66d4f
poetry install

# Run
rm -rf mutants # In case you already have stuff in the mutants directory
poetry run mutmut run
```

## Expected behaviour

Mutations are tested

## Actual behaviour

```
Exit code: 1 ➜ rm -rf mutants && poetry run mutmut run
...
Running forced fail test
...
=== 609 passed, 4 skipped in 23.40s ===
    exit code 0
FAILED
Exit code: 1 ➜ 
```

boxed:
Forced fail is a quick check to see if the mutants are being run. It's quite easy to get into a situation where the original code is first in the import list and then all mutants will survive. Which might be what happens in your case. During forced fail, ideally only one test should be run and that should immediately fail.

WillGibson:
I just updated the commit hash in the steps to replicate, it was the wrong one.

WillGibson:
Hi @boxed. I've forked our repository and brought this little challenge on holiday with me for the bits where `i just want to stay out of the midday sun ;-)

If you were to run the following commands, the problem manifests fairly quickly...

```
# Setup
git clone git@github.com:WillGibson/platform-tools.git
git checkout add-mutation-testing
poetry install
```

```
# Run
rm -rf mutants # In case you already have stuff in the mutants directory
poetry run mutmut run
```

[...]

PR #354, opened 2024-12-30
Improve error message when forced fail test fails

WillGibson:
I started with something small relating to https://github.com/boxed/mutmut/issues/349 in order to start familiarising myself with the Mutmut code/tests.


mutmut/__main__.py
```diff
@@ -885,14 +885,14 @@ def print_stats(source_file_mutation_data_by_path, force_output=False):
     print_status(f'{(s.total - s.not_checked)}/{s.total}  🎉 {s.killed} 🫥 {s.no_tests}  ⏰ {s.timeout}  🤔 {s.suspicious}  🙁 {s.survived}  🔇 {s.skipped}', force_output=force_output)
 
 
-def run_forced_fail(runner):
+def run_forced_fail_test(runner):
```
WillGibson, line 888: Renamed to better differentiate it from the `runner`'s `run_force_fail` method.

mutmut/__main__.py
```diff
@@ -885,14 +885,14 @@ def print_stats(source_file_mutation_data_by_path, force_output=False):
...
     os.environ['MUTANT_UNDER_TEST'] = 'fail'
     with CatchOutput(spinner_title='Running forced fail test') as catcher:
         try:
             if runner.run_forced_fail() == 0:
                 catcher.dump_output()
-                print("FAILED")
-                os._exit(1)
+                print("Error: Unable to force a test failure during the forced fail test")
+                raise SystemExit(1)
```
boxed, line 895: I would suggest "FAILED: Unable to force test failures". To be consistent, and for brevity.

Why did you change the exit? I use os._exit() for fast exit, so we don't spend time deallocating at the end of the process.

WillGibson: I have update the error message.

I switched to `SystemExit` because it is much easier to write a test for. Re the performance thing, I think this is something that will only happen once, early on in a run, if something has gone wrong, so some milliseconds taken occasionally doesn't seem like a problem.

WillGibson: Not sure why `test_run_forced_fail_test_with_all_tests_passing` ia passing locally, but not in GitHub Actions. I'll come back to that when I next have some time.

WillGibson: All the unit tests I added are failing in GitHub Actions because the output emitted by `run_forced_fail_test` is not making it to `stdout`.

I have tried adding some sleeps to rule out a race condition and changing the runner to MacOS in case that was a factor. No banana.

WillGibson:
@boxed I have poked around and set things up so that there is [a test that does not capture the output](https://github.com/boxed/mutmut/pull/354/files#diff-8ccb3dfbce788ac7fc87e69a419163b3bd5d2c31834c387c494999c6cfbd90a1R415) and running single tests with verbose etc.

If you look at this [GitHub Action run](https://github.com/WillGibson/mutmut/actions/runs/12739654893/job/35503659457), it seems that `CatchOutput` is only capturing the first line of the output in GitHub Actions for some reason. You can see "done" (and "FAILED" for come reason without the new text) in the test output.

My time for playing this morning has run out. If you have any thoughts on why this might be I would be interested in hearing them.

mutmut/__main__.py
WillGibson: @boxed I have poked around and set things up so that there is [a test that does not capture the output](https://github.com/boxed/mutmut/pull/354/files#diff-8ccb3dfbce788ac7fc87e69a419163b3bd5d2c31834c387c494999c6cfbd90a1R415) and running single tests with verbose etc.

If you look at this [GitHub Action run](https://github.com/WillGibson/mutmut/actions/runs/12739654893/job/35503659457), it seems that CatchOutput is only capturing the first line of the output in GitHub Actions for some reason. You can see "done" (and "FAILED" for come reason without the new text) in the test output.

My time for playing this morning has run out. If you have any thoughts on why this might be I would be interested in hearing them.

WillGibson: There is something different between MacOS and Ubuntu that means that last `print('    done')` is not in `capfd`'s `out` for the assert on Ubuntu.

Very odd.

WillGibson: Temporarily running the tests on MacOS as well as Ubuntu in GitHub Actions indicates it's not an OS difference.

I'm now wondering whether there is some difference in the way https://github.com/actions/setup-python installs Python compared to how I installed it (via asdf) on my laptop.

@boxed (or anyone else) If you checkout this branch and run the following, does it pass or fail, and how did you install Python?

```
python -m pytest -vvv --capture=no tests/test_mutation.py::test_run_forced_fail_test_with_failing_test
```

boxed: It passes

WillGibson: Thanks. But... damn. I wonder what the difference is in GitHub Actions.

WillGibson: If I comment out the `with CatchOutput` part in `run_forced_fail_test`, `capfd` works as expected.

I suspect `CatchOutput` and `capfd` are doing much the same thing behind the scenes and `CatchOutput` is usurping `capfd`. I'll need to read the source code for `capfd`, but boarding a plane imminently.

@boxed What is the intent/purpose of `CatchOutput`?

boxed: Yea that makes sense.

The purpose is to not show a lot of output when running mutmut, but being able to dump it on errors. Showing all output to the user would make the tool basically unusable.

WillGibson: @boxed I have not been able to get to the bottom of why `CatchOutput` does not play nicely with `capfd` in GitHub Actions, so I have negated the effects of CatchOutput for the purposes of these new tests by patching it.

Another option would be to change so that we pass `catcher` in, which would allow us to pass in a mock. I'm not sure if that's worth it right now.

I have rebased my six million commits and think this pull request is ready for review now.

Otto-AA:
Thanks for this PR, @WillGibson . I've tested it locally, looks good to me!

PR #355, opened 2024-12-30
Add initial contributing to Mutmut documentation

WillGibson:


tests/__init__.py
WillGibson, line 1: Added to stop the following error when running pytest...

```
ImportError while importing test module '/Users/will/Dev/WillGibson/mutmut/tests/test_mutation.py'.
Hint: make sure your test modules/packages have valid Python names.
```

CONTRIBUTING.rst
```diff
@@ -0,0 +1,35 @@
...
+
+.. code-block:: console
+
+    pip install -r <path_to_mutmut_codebase>/requirements.txt
+
+Run Mutmut:
+
+.. code-block:: console
+
+    python <path_to_mutmut_codebase>/mutmut/__main__.py run
```
boxed, line 35: I think it's more sane here to do `python -m pip --editable .` to install mutmut in the venv.

WillGibson: Good shout. I didn't know about `--editable`. I have update this section.

WillGibson:
@boxed Is this one approvable now?

PR #356, opened 2025-01-01
Test against all supported Python versions in GitHub Actions

WillGibson:
Includes a comment to save [questions like this](https://github.com/boxed/mutmut/commit/97e81e5634fd86de62e26a7de8ac640f9f1b15b6#commitcomment-150800695) people like me ;-)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,5 @@ click
 junit-xml==1.8
 setproctitle
 textual
+# We use the tomlib builtin for Python 3.11 and newer
 toml>=0.10.2; python_version < '3.11'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,16 +4,20 @@ on: [push, pull_request]
 
 jobs:
   tests:
-    name: Tests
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -3,4 +3,5 @@ click
 junit-xml==1.8
 setproctitle
 textual
+# We use the tomlib builtin for Python 3.11 and newer
```
WillGibson, line 6: Context: https://github.com/boxed/mutmut/commit/97e81e5634fd86de62e26a7de8ac640f9f1b15b6#commitcomment-150800695

boxed:
I guess we could drop py3.7.

WillGibson:
It passed on my fork...

https://github.com/WillGibson/mutmut/actions/runs/12566651166

But I think dropping support for end of lifed versions of Python is easy to make a case for.

WillGibson:
@boxed It looks like the `ubuntu-latest` tag went from 22 to 24 between our runs.

I have locked it to 22 for now so that we can separate decisions about which versions of Python to support from testing all the versions Mutmut says it supports.

Please can you approve the test run.

PR #372, opened 2025-03-17
copy full src dir before creating mutants

### LINKED ISSUE #371: also_copy overwrites mutants, run doesn't copy over non-python files in source folders
I am not sure what is more the issue. That the mutant generation only copies over python files or that the also_copy command overwrites the mutants since it runs afterwards.

Basically I have a repo with a bunch of schema definitions in JSON files and a few other files in the source directory. These files need to be present for the tests to run but when doing `mutmut run` it doesn't copy the files over since it only looks for python source files when copying them into the mutant folder and the tests fail. 

I could specify the dozens of schema files into also_copy but that is extremely tedious and prone to failure anytime a filename is changed or a new file is added. 

I think the solution would be to copy over the entire source directory then make the mutants and overwrite the source files with the new mutated files instead. I think that would solve both issues I am running into and just be a more resilient option. The `also_copy` config option could still exist to make sure files outside the source dir are included but only copying python source files is limiting.

DomAmato:
The current process only copies python files but tests might rely on other files that exist within the src directories in a repo. This PR copies the full src directories first, creates and overwrites the files with the mutants and then copies the files specified in `also_copy`

fixes #371

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -219,6 +219,13 @@ def _mutmut_trampoline(orig, mutants, *args, **kwargs):
 """
 yield_from_trampoline_impl = trampoline_impl.replace('result = ', 'result = yield from ').replace('_mutmut_trampoline', '_mutmut_yield_from_trampoline')
 
+def copy_src_dir():
+    for root, filename in walk_all_files():
+        path = Path(root) / filename
+        print(path)
+        output_path = Path('mutants') / path
+        makedirs(output_path.parent, exist_ok=True)
+        shutil.copy(path, output_path)
 
 def create_mutants():
     for path in walk_source_files():
@@ -1230,6 +1237,7 @@ def run(mutant_names, *, max_children):
     start = datetime.now()
     makedirs(Path('mutants'), exist_ok=True)
     with CatchOutput(spinner_title='Generating mutants'):
+        copy_src_dir()
         create_mutants()
         copy_also_copy_files()
 
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -261,10 +261,6 @@ def pragma_no_mutate_lines(source):
 def create_mutants_for_file(filename, output_path):
     input_stat = os.stat(filename)
 
-    # if output_path.exists() and output_path.stat().st_mtime == input_stat.st_mtime:
-    #     # print('    skipped', output_path, 'already up to date')
-    #     return
-
     with open(filename) as f:
         source = f.read()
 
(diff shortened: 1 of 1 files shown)


mutmut/__main__.py
```diff
@@ -219,6 +219,13 @@ def _mutmut_trampoline(orig, mutants, *args, **kwargs):
 """
 yield_from_trampoline_impl = trampoline_impl.replace('result = ', 'result = yield from ').replace('_mutmut_trampoline', '_mutmut_yield_from_trampoline')
 
+def copy_src_dir():
+    for root, filename in walk_all_files():
+        path = Path(root) / filename
+        print(path)
```
boxed, line 225: The print seems like it's accidentally left over?

mutmut/__main__.py
```diff
@@ -219,6 +219,13 @@ def _mutmut_trampoline(orig, mutants, *args, **kwargs):
 """
 yield_from_trampoline_impl = trampoline_impl.replace('result = ', 'result = yield from ').replace('_mutmut_trampoline', '_mutmut_yield_from_trampoline')
 
+def copy_src_dir():
```
boxed, line 222: This entire function could be just `shutil.copytree` no?

DomAmato: I thought so too but when I rewrote it it was not working as expected. I'll rework it and test it again because that should be much faster

DomAmato: Ah I see why I was running into that issue.

```
def create_mutants_for_file(filename, output_path):
    input_stat = os.stat(filename)

    if output_path.exists() and output_path.stat().st_mtime == input_stat.st_mtime:
        # print('    skipped', output_path, 'already up to date')
        return
```

It was triggering this conditional and not creating the mutated files

DomAmato: commenting it out works as expected now, thoughts?

boxed: Looks more reasonable yea. If you remove the commented out code I can merge it.

mutmut/__main__.py
```diff
@@ -1370,7 +1375,8 @@ def read_one_child_exit_status():
 
     print_stats(source_file_mutation_data_by_path, force_output=True)
     print()
-    print(f'{count_tried / t.total_seconds():.2f} mutations/second')
+    print(f'    done in {round(t.total_seconds()*1000)}ms')
+    print(f'    {count_tried / t.total_seconds():.2f} mutations/second')
```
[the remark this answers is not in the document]
DomAmato: can remove, just wanted to understand how long the tests took to run

mutmut/__main__.py
```diff
@@ -257,10 +261,6 @@ def pragma_no_mutate_lines(source):
 def create_mutants_for_file(filename, output_path):
     input_stat = os.stat(filename)
 
-    if output_path.exists() and output_path.stat().st_mtime == input_stat.st_mtime:
```
boxed, line 260: What is this deletion? This PR should be about just copying the file tree as the title says.

DomAmato: I explained it in a previous comment but if that code is there it doesn't create the mutants because `output_path.stat().st_mtime == input_stat.st_mtime` will be true so either this logic needs to be modified or it removed for it to work with the `shutil.copytree`

boxed: Ah. My bad.

boxed:
Thanks!

PR #377, opened 2025-04-10
Add --show-killed option for mutmut browse

Otto-AA:
Now running `mutmut browse --show-killed` will also display killed mutants in the UI. This can be helpful for debugging, and maybe also as a user to understand what `mutmut` does.


boxed:
Nice

Issue #370, opened 2025-03-07
"FAILED" after "forced fail test"


MarcoGorelli:
I'm trying to run mutmut on the Narwhals project, and have set
```
[tool.mutmut]
paths_to_mutate = [ "narwhals/_arrow/series.py" ]
tests_dir = [ "tests/expr_and_series/cum_sum_test.py" ]
do_not_mutate= ["tests/utils.py", "narwhals/stable/v1/__init__.py"]
```
in pyproject.toml

I then run `mutmut run` and get
```
⠇ Generating mutants
    done in 1068ms
⠧ Running stats     
    done
⠦ Running clean tests
    done
⠴ Running forced fail test...................................................................................................................................................................................s..s [ 12%]
.................................................xxxxxx.................................................................................x............s................................. [ 25%]
...................s..s..........................................................................................s..................................................................... [ 38%]
............x....xxxxxxxs...................x..........................................sss.............x...x.x...x...x.....x......x..x....................s......................xxx..x [ 50%]
xxx.xxx..xx............s...............................................................................................................................................s............... [ 63%]
..........x.....x.xx....x...........x....x...............................x..................s................................................................................x......... [ 76%]
..............xxxxxxxxxxxxxxxxxxxxxxxxx.....................................s...................................................s...........s...............x....xx.................... [ 89%]
............................................................................s.....................................................ss................s.......                            [100%]
1343 passed, 21 skipped, 73 xfailed in 4.79s
FAILED
```

I don't understand the output, nor what I'm supposed to do

If I run `mutmut browse` then it shows 1295 with `?` status. I don't know how to interpret that

Any help making sense of how to use mutmut would be appreciated 🙏

boxed:
The forced fail test should fail immediately. So that any tests are running means it doesn't work. The idea with the forced fail is that all mutants are set to fail so that we can check that mutants can be triggered at all. This is the part that didn't work. 

Still.. that some tests fail seems weird too..

alfmorais:
Hello from Brazil,

I'm experiencing the same issue.

My test suite has 47 unit tests. When I run the command:

```bash
mutmut run
```

The output is:

```bash
(project-service-py3.12) alfredo@machine-address:~/Company/project-service$ mutmut run
⠹ Generating mutants
    done in 12ms
⠧ Running stats     
    done
⠹ Running clean tests
    done
⠧ Running forced fail test
****pytest output***
PASSED                                                                                                                   [100%]

[...]

PR #379, opened 2025-04-12
Remove non-mutated code from sys.path

Otto-AA:
This change should help troubleshooting errors, where the package setup in `./mutants` is not correct.

Currently, when the tests in `./mutants/` try to import a package that is not in `./mutants/`, it will also look in the original (non-mutated) source code. Is there ever a case, where this is desired? I would guess that `also_copy` covers these use cases, but I am not sure, so I wanted to ask for your opinion @boxed .

The reason I want to prevent importing the original code is that it hides the error cause if the package setup of `./mutants` is wrong. If the package is not setup properly (e.g. #370 , which only has a single file for `paths_to_mutate` rather than the whole source), the tests currently just use the original code and thus always pass. We only stop at the forced fail test.

With these changes, it will already fail at the test collecting stage. If debug=true, we also have the error cause reported:

```bash
ImportError while loading conftest '/.../narwhals/mutants/tests/conftest.py'.
../tests/conftest.py:14: in <module>
    from narwhals.utils import generate_temporary_column_name
E   ModuleNotFoundError: No module named 'narwhals.utils'
    exit code 4
```

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -170,7 +170,9 @@ class CollectTestsFailedException(Exception):
 
 
 class BadTestExecutionCommandsException(Exception):
-    pass
+    def __init__(self, pytest_args: list[str]) -> None:
+        msg = f'Failed to run pytest with args: {pytest_args}. If your config sets debug=true, the original pytest error should be above.'
+        super().__init__(msg)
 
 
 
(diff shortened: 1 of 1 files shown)


boxed:
I like it!

PR #380, opened 2025-04-20
Expand mutation operators with built‑in name swaps and regex fuzzing

Magic-Man-us:
- Extend operator_name to flip more built‑ins (len↔sum, min↔max, all↔any, str↔repr, etc.)
- Add operator_regex to mutate regex literals (quantifier swaps, {1,}↔+, \d↔[0-9], char‐class reversal)
- Register operator_regex alongside other call mutations
- Update tests to cover new regex and name‐swap behaviors

### DIFF
--- a/tests/test_mutation.py
+++ b/tests/test_mutation.py
@@ -9,7 +9,6 @@ from mutmut.__main__ import (
     get_diff_for_mutant,
     orig_function_and_class_names_from_key,
     run_forced_fail_test,
-    Config,
     MutmutProgrammaticFailException,
     CatchOutput,
 )
@@ -513,7 +512,7 @@ def test_run_forced_fail_test_with_failing_test(_start, _stop, _dump_output, cap
     print(f"out: {out}")
     print(f"err: {err}")
     assert 'done' in out
-    assert os.environ['MUTANT_UNDER_TEST'] is ''
+    assert os.environ['MUTANT_UNDER_TEST'] == ''
 
 
 # Negate the effects of CatchOutput because it does not play nicely with capfd in GitHub Actions
@@ -528,7 +527,7 @@ def test_run_forced_fail_test_with_mutmut_programmatic_fail_exception(_start, _s
 
     out, err = capfd.readouterr()
     assert 'done' in out
-    assert os.environ['MUTANT_UNDER_TEST'] is ''
+    assert os.environ['MUTANT_UNDER_TEST'] == ''
 
 
 # Negate the effects of CatchOutput because it does not play nicely with capfd in GitHub Actions
@@ -542,7 +541,7 @@ def test_run_forced_fail_test_with_all_tests_passing(_start, _stop, _dump_output
     with pytest.raises(SystemExit) as error:
         run_forced_fail_test(runner)
 
-    assert error.value.code is 1
+    assert error.value.code == 1
     out, err = capfd.readouterr()
     assert 'FAILED: Unable to force test failures' in out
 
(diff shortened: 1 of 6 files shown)


Otto-AA:
Hi, thanks your contribution, many new mutations that could uncover missing tests :)

There are some mutations, where to me it seems they don't bring much value to mutation testing. If you want, some of them could be reworked, or you can simply drop them for now and we can revisit them later.

In particular, some of the mutations seem like "[equivalent mutants](https://stryker-mutator.io/docs/mutation-testing-elements/equivalent-mutants/)", I think we should remove or adapt them. When we would include these mutations, the unit tests won't have a chance to detect them as the mutation behaves identical to the original code. So these are like false positives and we should try to keep their number low.

The following mutations seem like they could produce many equivalent mutants:
- `str <-> repr`: in many cases these are equivalent, e.g. for builtin numbers, or also for attrs classes. There are cases where this is not the case, but I feel like it's not worth it.
- `list <-> tuple`: *If* the program does not raise a runtime exception because of e.g. calling `.append` on a tuple, then I think these two behave similar. And if it throws a runtime exception, we only check code coverage of this line (where tools like `coverage.py` are more suited)
- `set <-> frozenset`: same as for `list <-> tuple`; in the cases where no exception is raised, they likely behave identical
- `{1,} -> +`: From my regex understanding, these are identical, and thus indistinguishable for the tests

The `chr <-> ord` swap also always (?) raises an Exception. For this, we could map `chr(123) -> chr(123 + 1)` and `ord(c) -> ord(c) + 1`, to check if changing the result slightly makes any difference.

Regarding `len <-> sum` and `map <-> filter`, I think these are convenient choices that fit into our current name mapping. If we want to spend some more time on the implementation, I would suggest to map `len(...) -> len(...) + 1`, `sum(...) -> sum(...) + 1` and `map(fn, arr) -> list(arr)`, `filter(fn, arr) -> list(arr)`.

tests/test_mutation.py
```diff
@@ -667,3 +666,40 @@ def add(self, *args, **kwargs):
...
+])
+def test_regex_mutations_loose(original, want_patterns):
+    mutants = mutants_for_source(original)
+    # remove the generic "foo(None)" from operator_arg_removal
+    # and the SimpleString "XX…" from operator_string
+    filtered = [
+        m for m in mutants
+        if "(None)" not in m 
+        and "XX" not in m
+    ]
```
Otto-AA, line 702: This filtering seems redundant to me. If `want in filtered` then `want in mutants` is also true.

Magic-Man-us: You are right, I will remove this

Magic-Man-us:
You are spot on with a lot of this. I apologize for not getting back earlier its been a busy week .
I will remove the equal equivalents, or perhaps do a +1. 


In the meantime I was thinking of some problems I saw in the past and of course, I did not experience  😅 ...
One really good candidate I will add to the list is Enum to Str etc.  it should cause string comparisons will fail silently >:) 
JSON serialization or API payloads will break if not properly handled.
and Database operations or other logic expecting strings or ints will raise exceptions or errors.
Also Enum to Int!

I will update sometime this weekend. 
Thank you for the feedback @Otto-AA

Otto-AA:
Hi, you can ping me when your changes are review-ready :)
(Though I will probably need some days to find the time for reviewing)

Magic-Man-us:
Thanks for the feedback! I've addressed your concerns about equivalent mutants.

This introduces improvements to the mutation testing focusing on eliminating equivalent mutations as you had mentioned prior, additionally, adding meaningful function call mutations, and enhancing regex mutation handling. it includes comprehensive updates to tests for better coverage and accuracy. Below is a summary of the most important changes grouped by theme.

### Mutation Framework Enhancements

* Eliminated equivalent mutations for common operations (`len ↔ sum`, `min ↔ max`, etc.) to reduce false positives and improve mutation score accuracy. [[1]](diffhunk://#diff-21d5c998dddab8f71a9e1271172994663897d95ef21434ded062d1379c6520e7R1-R81) [[2]](diffhunk://#diff-f8f856b379d812f594849e512fc044972e8b12563d962ffc3e881b3bd8e6726bL155-R176)
* Added new function call mutations for aggregate functions (`len(...) → len(...) ± 1`) and mapping/filtering functions (`map(fn, arr) → list(arr)`), ensuring meaningful semantic changes. [[1]](diffhunk://#diff-21d5c998dddab8f71a9e1271172994663897d95ef21434ded062d1379c6520e7R1-R81) [[2]](diffhunk://#diff-f8f856b379d812f594849e512fc044972e8b12563d962ffc3e881b3bd8e6726bR437-R440)
* Improved regex mutation handling by avoiding equivalent mutations (`{1,} ↔ +`) and introducing meaningful alternatives (`{1,} → {2,}` or `{0,}`). [[1]](diffhunk://#diff-21d5c998dddab8f71a9e1271172994663897d95ef21434ded062d1379c6520e7R1-R81) [[2]](diffhunk://#diff-f8f856b379d812f594849e512fc044972e8b12563d962ffc3e881b3bd8e6726bR251-R424)

### Code Quality Improvements

* Corrected typos in comments and documentation, such as fixing "indidvidual" to "individual."
* Preserved function signature integrity in `chr/ord` mutations to avoid runtime exceptions, ensuring safer mutations. [[1]](diffhunk://#diff-21d5c998dddab8f71a9e1271172994663897d95ef21434ded062d1379c6520e7R1-R81) [[2]](diffhunk://#diff-f8f856b379d812f594849e512fc044972e8b12563d962ffc3e881b3bd8e6726bR437-R440)

### Test Coverage Enhancements

* Added extensive parameterized tests for regex mutations to ensure complete coverage of mutation patterns and edge cases.
* Introduced comprehensive tests for subscript mutations, including edge cases like slices with non-integer bounds and tuple subscripts.
* Updated assertions in existing tests to use `==` instead of `is` for string and integer comparisons, improving correctness. [[1]](diffhunk://#diff-8ccb3dfbce788ac7fc87e69a419163b3bd5d2c31834c387c494999c6cfbd90a1L516-R515) [[2]](diffhunk://#diff-8ccb3dfbce788ac7fc87e69a419163b3bd5d2c31834c387c494999c6cfbd90a1L545-R544)

These changes collectively enhance the reliability and effectiveness of the mutation testing framework while ensuring robust test coverage.

Issue #297, opened 2023-12-14
Mutate str methods?


jimrybarski:
Is there any interest in adding support for mutating methods on str objects? For example, `s.lower()` would be mutated to `s.upper()`, or `s.strip()` would be mutated to `s`. Of course, not all methods have a clear mutation (e.g. `str.maketrans()`) but I think enough of them do that even partial coverage could be useful.

CHB-0r1s:
I guess this issue resolved after last pr))
@Otto-AA @boxed

Otto-AA:
Yes, implemented by #386

PR #384, opened 2025-05-08
Add .upper() .lower() and .capitalize() as mutations

CHB-0r1s:
My first small pr into this great project :)

Btw edit tests to cover all new mutations and edit e2e


CHB-0r1s:
this pr connected with https://github.com/boxed/mutmut/issues/297

but idk how to connect pr and issue))

CHB-0r1s:
I'm not completely sure... the last commit is ok? It can cause inconsistency, when one type of mutations can "do nothing", rather others "can't do nothing".

mutmut/node_mutation.py
```diff
@@ -37,7 +37,17 @@ def operator_string(
...
-        yield node.with_changes(value=f"{prefix}{value[0]}XX{value[1:-1]}XX{value[-1]}")
+        supported_str_mutations: list[Callable[[str], str]] = [
+            lambda x: "XX" + x + "XX",
+            lambda x: x.lower(),
+            lambda x: x.upper(),
+            lambda x: x.capitalize(),
+        ]
+
+        for mut_func in supported_str_mutations:
+            if mut_func(value[1:-1]) == value[1:-1]: # because mutant which do nothing is useless
```
boxed, line 48: This line looks wrong. What is that supposed to do?

CHB-0r1s: Thank u for your comment. I want to prevent cases when mutations do not modify original string.

For example, cases when 'foo' (original string) == 'foo'.lower(). There we have 'useless mutant'. That's the point.

CHB-0r1s: I this approach is valid, I can modify e2e test_cases for it

CHB-0r1s: I just don't know, is this approach good

boxed: Hmm, yea ok. I see it now. Not very clear. How about
```py
new_value = f"{prefix}{value[0]}{mut_func(value[1:-1])}{value[-1]}"
if new_value == value:
    continue
yield node.with_changes(value=new_value)
```
that reads much clearer to me.

CHB-0r1s: Ok, sounds fair

CHB-0r1s: fix that and also fix e2e

CHB-0r1s: @boxed 
Do you want me to fix something else?

boxed:
Thanks!

mutmut/node_mutation.py
boxed: No, that's much clearer. I'm happy with this.

boxed:
#297 is a slightly different thing, that's about mutating calls, this PR is about mutating string literals.

PR #386, opened 2025-05-09
Add str methods call mutations and tests

CHB-0r1s:
Add string methods as:
- lower/upper
- lstrip/rstrip
- find/rfind
- ljust/rjust
- index/rindex
- split/rsplit
- removeprefix/removesuffix
- partition/rpartition

### DIFF
--- a/mutmut/node_mutation.py
+++ b/mutmut/node_mutation.py
@@ -122,7 +122,8 @@ def operator_string_methods_swap(
 
      for old_call, new_call in supported_str_methods_swap:
          if m.matches(node.func, m.Attribute(value=m.DoNotCare(),  attr=m.Name(value=old_call))):
-             yield node.with_changes(func=cst.Attribute(value=node.func.value, attr=cst.Name(value=new_call)))
+            func_name = cst.ensure_type(node.func, cst.Attribute).attr
+            yield node.with_deep_changes(func_name, value=new_call)
 
 
 def operator_remove_unary_ops(
--- a/mutmut/node_mutation.py
+++ b/mutmut/node_mutation.py
@@ -96,6 +96,35 @@ def operator_arg_removal(
             yield node.with_changes(args=[*node.args[:i], *node.args[i + 1 :]])
 
 
+supported_str_methods_swap = [
+         ("lower", "upper"),
+         ("upper", "lower"),
+         ("lstrip", "rstrip"),
+         ("rstrip", "lstrip"),
+         ("find", "rfind"),
+         ("rfind", "find"),
+         ("ljust", "rjust"),
+         ("rjust", "ljust"),
+         ("index", "rindex"),
+         ("rindex", "index"),
+         ("split", "rsplit"),
+         ("rsplit", "split"),
+         ("removeprefix", "removesuffix"),
+         ("removesuffix", "removeprefix"),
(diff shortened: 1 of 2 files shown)


CHB-0r1s:
@boxed @Otto-AA 
Now the pr is really about this issue
https://github.com/boxed/mutmut/issues/297

CHB-0r1s:
If it necessary I can fix last test to cover .upper() test case too, but more indexes (such as x_capitalize__mutmut_1, x_capitalize__mutmut_2, ..., x_capitalize__mutmut_n) can increase test complexity

tests/test_mutation.py
```diff
@@ -40,6 +40,32 @@ def mutated_module(source: str) -> str:
         # ('break', 'continue'),  # probably a bad idea. Can introduce infinite loops.
         ('break', 'return'),
         ('continue', 'break'),
+        ('a.lower()', 'a.upper()'),
+        ('a.upper()', 'a.lower()'),
```
Otto-AA, line 44: Could you add a test for `('a.b.lower()', 'a.b.upper()'),` here? Your code already correctly handles this case, but would be nice to cover in the tests.

```suggestion
        ('a.upper()', 'a.lower()'),
        ('a.b.lower()', 'a.b.upper()'),
```

mutmut/node_mutation.py
```diff
@@ -96,6 +96,35 @@ def operator_arg_removal(
...
+     ]
+
+def operator_string_methods_swap(
+     node: cst.Call
+ ) -> Iterable[cst.Call]:
+     """try to swap string method to opposite e.g. a.lower() -> a.upper()"""
+
+     for old_call, new_call in supported_str_methods_swap:
+         if m.matches(node.func, m.Attribute(value=m.DoNotCare(),  attr=m.Name(value=old_call))):
+             yield node.with_changes(func=cst.Attribute(value=node.func.value, attr=cst.Name(value=new_call)))
```
Otto-AA, line 125: I'd prefer to use `with_changes` (or `with_deep_changes`) to only modify the function name and cloning the rest as it was before. This makes sure that whitespace and other stuff belonging to the attribute is copied, and we only change the function name.

```suggestion
            func_name = cst.ensure_type(node.func, cst.Attribute).attr
            yield node.with_deep_changes(func_name, value=new_call)
```

Otto-AA (commented):
Thanks, looks good to me!

I've only added some small remarks.

mutmut/node_mutation.py
CHB-0r1s: Thank you a lot, that my first time with libcst! I definitely agree with your preferences, it's much better

tests/test_mutation.py
CHB-0r1s: Thanks for the suggestion! It's a good idea to have explicit test coverage for nested attribute access.

CHB-0r1s: @Otto-AA 
Tests were added)
Do you want me to fix something else?

boxed:
Nice. @Otto-AA feel free to merge this when you are satisfied with it.

mutmut/node_mutation.py
Otto-AA: You're welcome :) Thank you for your contribution!

PR #388, opened 2025-05-23
Fix string mutation

fabi321:
String mutation would otherwise fail when escape sequences are used,
changing `"\n"` to `"\N"`, which is a syntax error.

### DIFF
--- a/tests/test_mutation.py
+++ b/tests/test_mutation.py
@@ -87,6 +87,7 @@ def mutated_module(source: str) -> str:
         ('a: Optional[int] = None', 'a: Optional[int] = ""'),
         ('a: int = 1', ['a: int = 2', 'a: int = None']),
         ('a: str = "FoO"', ['a: str = "XXFoOXX"', 'a: str = "foo"', 'a: str = "FOO"', 'a: str = "Foo"', 'a: str = None']),
+        ('a: str = "Fo\\t"', ['a: str = "XXFo\\tXX"', 'a: str = "fo\\t"', 'a: str = "FO\\t"', 'a: str = None']),
         ('lambda: 0', ['lambda: 1', 'lambda: None']),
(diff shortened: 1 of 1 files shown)


Otto-AA:
I think following regex would work to ignore the escape sequences:
```python
NON_ESCAPE_SEQUENCE = re.compile(r"((?<!\\)[^\\]+)")
NON_ESCAPE_SEQUENCE.sub(lambda match: match.group(1).upper(), s)
```

> Or not touch any string with escape sequences, but that feels like a lost opportunity.

We are already modifying the string with the XX+value+XX modification. This should cover many cases already. We might miss some cases where the string capitalization does not matter for the program behaviour (e.g. SELECT vs select in SQL), so if someone has no tests for this it could remain unnoticed by mutmut, but that would also be fine imo.

fabi321:
That regex only works in most cases, as it would not convert `"\\\\n"` to upper case.
As you already mentioned that there are other modifications to account for this, I believe it's best not to touch any string containing escape sequences, neither with `upper()`, nor `lower()` or `title()`.

Otto-AA:
> That regex only works in most cases, as it would not convert "\\\\n" to upper case.

Yes, I think the edge cases are always on the safe side and do not produce a syntax error. If I am not mistaken, it creates a working mutation in most cases, and in other cases it creates no invalid mutation (even converting "\N{grinning face}" to "\N{GRINNING FACE}" is on the safe side). If that is true, I think we should use it and add a method with a lot of escape sequences to an e2e test.

Otto-AA:
Hi @fabi321 , are you still interested in working on this? If not, I would I would test out the ` NON_ESCAPE_SEQUENCE` I proposed above and if it works merge it this way. Thank you either way for your contribution :heart:

Issue #389, opened 2025-05-23
Whitelist rsplit


fabi321:
`str.rsplit` behaves exactly like `str.split`, if maxsplit is not provided, so in that case there should be no mutation.

Otto-AA:
> str.rsplit behaves exactly like str.split, if maxsplit is not provided, so in that case there should be no mutation.

Yes, without the maxsplit argument this mutation cannot be detected by the tests. We should change not mutate in this case. If you want to take a shot at this, I would suggest to add another `if m.matches(...)` to the [operator_string_methods_swap](https://github.com/boxed/mutmut/blob/b124c6ad46b0a9a048b5e47c8b7f93d33a9f2e6a/mutmut/node_mutation.py#L118) method (outside of the loop) and remove [this](https://github.com/boxed/mutmut/blob/b124c6ad46b0a9a048b5e47c8b7f93d33a9f2e6a/mutmut/node_mutation.py#L110C10-L110C30) dict entry.

> Upon looking mutations, I also fund another false positive: I have assert False in my code, in order to cover branches that are not intended to be hit. Changing them to true doesn't make much sense.
It also tried to mutate the string of the assert, which also does not make too much sense.

I also do not think mutating `assert false` is useful. We could change this, but I think the change would not be trivial. Currently most (or all?) mutation operators do not rely on the context, so the mutation of `False` to `True` currently does not know if it is the operand of an `assert` statement. We could considert to add the [parent metadata](https://libcst.readthedocs.io/en/latest/metadata.html#parent-node-metadata) as additional context for operators.

So I'm not sure if this would be worth the effort. The mutation for `assert False` should only show up as a mutation that was not executed (similar to uncovered code with e.g. coverage.py).

> In general, I don't feel like it's that useful to test for mutations in asserts, as they are mostly a defensive measure (they can be disabled, after all).

I guess people could use `assert` statements for quick and dirty input validation, for instance:

```python
def save_user(username, age):
  assert 0 <= age <= 150, 'Please provide an age between 0 and 150'
```

While this is not best practice, I could imagine code like this. For these cases, mutating `assert` statements seems reasonable to me. If this is really uncommon, we could consider not mutating `assert` statements, which would be rather simple to implement in mutmut.

fabi321:
> So I'm not sure if this would be worth the effort. The mutation for `assert False` should only show up as a mutation that was not executed (similar to uncovered code with e.g. coverage.py).

That assertion comes up as fail. I noticed it due to that.

[...]

PR #394, opened 2025-06-14
Fix split / rsplit collision

CHB-0r1s:
After research of all possible combinations, there is such schema for prevent useless mutations:
###  Without mutations
1) split() -> split()
2) rsplit() -> rsplit()
3) split(" ") -> split(" ")
4) rsplit(" ") -> rsplit(" ")
5) split(sep=" ") -> split(sep=" ")
6) rsplit(sep=" ") -> rsplit(sep=" ")
7) split(maxsplit=-1) -> split(maxsplit=-1)
8) rsplit(maxsplit=-1) -> rsplit(maxsplit=-1)
9) split(" ", maxsplit=-1) -> split(" ", maxsplit=-1)
10) rsplit(" ", maxsplit=-1) -> rsplit(" ", maxsplit=-1)

###  With mutations
1) split(maxsplit=1) -> rsplit(maxsplit=1)
2) rsplit(maxsplit=1) -> split(maxsplit=1)
3) split(" ", 1) -> rsplit(" ", 1)
4) rsplit(" ", 1) -> split(" ", 1)
5) split(" ", maxsplit=1) -> rsplit(" ", maxsplit=1)
6) rsplit(" ", maxsplit=1) -> split(" ", maxsplit=1)

I save it as a note in comments.

### DIFF
--- a/mutmut/node_mutation.py
+++ b/mutmut/node_mutation.py
@@ -107,13 +107,16 @@ supported_str_methods_swap = [
          ("rjust", "ljust"),
          ("index", "rindex"),
          ("rindex", "index"),
-         ("split", "rsplit"),
-         ("rsplit", "split"),
          ("removeprefix", "removesuffix"),
          ("removesuffix", "removeprefix"),
          ("partition", "rpartition"),
          ("rpartition", "partition")
-     ]
+]
+
+supported_unsymmetrical_str_methods_swap = [
+    ("split", "rsplit"),
+    ("rsplit", "split")
+]
 
 def operator_string_methods_swap(
      node: cst.Call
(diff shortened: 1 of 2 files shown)


CHB-0r1s:
I hope this fix can help with https://github.com/boxed/mutmut/issues/389

CHB-0r1s:
@Otto-AA @boxed This magic with parsing -1 and +1 in libcst parser cause a lot of troubles for me. So, please, if you know better way to do it, I appreciate your help.

Otto-AA:
Hi, thank you for this PR!

> @Otto-AA @boxed This magic with parsing -1 and +1 in libcst parser cause a lot of troubles for me. So, please, if you know better way to do it, I appreciate your help.

I don't think that ` x.split(maxsplit=-1)` is written directly in the code. This is equivalent to ` x.split()`, which I would expect people to use, rather than explicitly providing and disabling `maxsplit`.

So I would suggest that we do not check if the value is `-1` and simply always mutate, if `maxsplit` is provided with any value. So `x.split(' ', -1)` would also be mutated to `x.rsplit(' ', -1)`, which is not optimal but should happen very seldomly.

mutmut/node_mutation.py
```diff
@@ -125,6 +128,62 @@ def operator_string_methods_swap(
...
+                        not len(node.args) or
+                        (len(node.args) == 1 and "maxsplit" not in key_args) or
+                        ("maxsplit" in key_args and maxsplit_val == -1)
+                ):
+                    continue
+
+                if (
+                        (len(node.args) == 2 and "maxsplit" not in key_args) or
+                        ("maxsplit" in key_args and maxsplit_val != -1)
+                ):
```
Otto-AA, line 182: We could rewrite it with the libcst matchers. I think both ways are quite readable, so choose whatever you prefer.

This would be the option with the libcst methods:

```python
                # either maxsplit is passed as the second arg (so 2+ args)
                if (m.matches(node, m.Call(args=[m.AtLeastN(n=2)]))
                    # or it is passed at any possition as a keyword arg
                    or m.matches(node, m.Call(args=[m.ZeroOrMore(), m.Arg(keyword=m.Name(value="maxsplit")), m.ZeroOrMore()]))):

                    func_name = cst.ensure_type(node.func, cst.Attribute).attr
                    yield node.with_deep_changes(func_name, value=new_call)
```

mutmut/node_mutation.py
```diff
@@ -125,6 +128,62 @@ def operator_string_methods_swap(
...
+
+                if (
+                        not len(node.args) or
+                        (len(node.args) == 1 and "maxsplit" not in key_args) or
+                        ("maxsplit" in key_args and maxsplit_val == -1)
+                ):
+                    continue
+
+                if (
+                        (len(node.args) == 2 and "maxsplit" not in key_args) or
```
Otto-AA, line 180: It cannot happen that we have two args *and* `maxsplit` as a keyword arg (we cannot pass the argument both as a positional argument and keyword argument). So we could remove the ` and "maxsplit" not in key_args` part here.

mutmut/node_mutation.py
```diff
@@ -125,6 +128,62 @@ def operator_string_methods_swap(
             func_name = cst.ensure_type(node.func, cst.Attribute).attr
             yield node.with_deep_changes(func_name, value=new_call)
 
+def unsymmetrical_string_methods_swap(
+    node: cst.Call
+) -> Iterable[cst.Call]:
+    """Try to handle specific mutations of string, which useful only in specific args combination."""
+
+    for old_call, new_call in supported_unsymmetrical_str_methods_swap:
+        if m.matches(node.func, m.Attribute(value=m.DoNotCare(), attr=m.Name(value=old_call))):
```
Otto-AA, line 137: libcst does not verify the properties that we do not mention, so the `value=m.DoNotCare()` is superfluous.

```suggestion
        if m.matches(node.func, m.Attribute(attr=m.Name(value=old_call))):
```

CHB-0r1s:
Thank you @Otto-AA ! 
I will apply your suggestions as soon as possible today)

CHB-0r1s:
So, new schema

### Without mutations
1. split() -> split()
2. rsplit() -> rsplit()
3. split(" ") -> split(" ")
4. rsplit(" ") -> rsplit(" ")
5. split(sep=" ") -> split(sep=" ")
6. rsplit(sep=" ") -> rsplit(sep=" ")

### With mutations
1. split(" ", 1) -> rsplit(" ", 1)
2. rsplit(" ", 1) -> split(" ", 1)
3. split(maxsplit=1) -> rsplit(maxsplit=1)
4. rsplit(maxsplit=1) -> split(maxsplit=1)
5. split(maxsplit=-1) -> rsplit(maxsplit=-1)
6. rsplit(maxsplit=-1) -> split(maxsplit=-1)
7. split(" ", maxsplit=-1) -> rsplit(" ", maxsplit=-1)
8. rsplit(" ", maxsplit=-1) -> split(" ", maxsplit=-1)
9. split(" ", maxsplit=1) -> rsplit(" ", maxsplit=1)
10. rsplit(" ", maxsplit=1) -> split(" ", maxsplit=1)

CHB-0r1s:
Hello! I applied all suggestions. Code looks pretty easy now. Schema was removed from comments (add link).
Can you check new updates, please :3

@Otto-AA @boxed

CHB-0r1s:
Just little ping @Otto-AA

PR #401, opened 2025-07-01
Document installation requirements (#400)

### LINKED ISSUE #400: Document that installation requires rustc and cargo
Hi there,

Thank you for this tool ! I found that in order to install it (and in particular to build the libcst dependency), it is necessary to have cargo and rustc required. This is not described in the README...

jedesroches:
closes #400


boxed:
The change suggested is too narrow. It seems to me like it could be "if you get error X, install cargo".

PR #404, opened 2025-07-06
Windows support

### LINKED ISSUE #397: Windows support
Mutmut 3 currently can be run on Windows through the WSL, but not natively on Windows.

We use `fork()` as a performance optimization, which only works on Unix systems. However, it should be possible to rewrite mutmut to also work with other multiprocessing options. This would be slower (as we cannot copy the whole memory at fork time, including the initialized pytest setup), but still a nice to have.

If someone wants to add native Windows support, we likely need to:
- rewrite the `_run()` method to use a the `spawn` method instead of `os.fork()`
- pass all necessary arguments to this process (currently we rely on global + local variables to be copied via `fork`ing the whole process memory)

I would consider following approaches, though there may be better ones:
- create a new process per mutant (as we do now)
- create `max_children` worker processes and initialize pytest there. The worker processes keep polling mutation tasks from a Queue and perform them
- use a `multiprocessing.pool.Pool` of size `max_children` and initialize pytest in the setup

While doing so, we need to be careful that the `fork()` version supports the same features and is as performant as before. In particular, we should support all possible test errors (e.g. what happens when a unit test calls `os.exit()` or `os._exit()`? What happens if the test runs into an infinite loop? If we currently support it with the `fork()` method, then we must also support it after the refactoring).

Otto-AA:
Refactor the code to also support windows. Closes #397 

This is still a WIP. If anyone wants to finish this PR, feel free to leave a comment and take it over :)
I probably won't get much free time in the next weeks, but wanted to share the current status.

## Implementation idea

The main tasks to support Windows are (1) to support the `spawn` method and (2) to support timeouts properly (if possible).

When directly using the `spawn` method it is much slower, because we need to startup everything from scratch for each mutant. This includes the python interpreter (I think) and the whole pytest test collection (ie importing the system under test).

To improve this performance, we only create `max_children` processes as "workers". These workers infinitely loop on:
1. pick the next mutant from a shared queue
2. test the mutant
3. return the result
4. repeat at (1) until the queue is empty

As such we only need to setup a limited number of processes, instead of one per mutant. To support timeouts (where we likely still want to kill processes), we need to restart workers when they time out.

## TODO

- [x] support `spawn` method by passing all globals as args to the new processes
- [x] use a process pool to increase performance (when we reuse the processes, pytest caches all the imported files)
- [x] add Windows tests to CI
- [ ] Reimplement timeouts (I've removed them for now for simplicity)
- [ ] Debug why this branch outputs slightly different results than on main
- [ ] Test on other repos
- [ ] Cleanup

### Implementing timeouts

Previously we used `resource` to implement timeouts: https://github.com/boxed/mutmut/blob/f63029b239fa7c8042b2a45c5191674df419cd31/mutmut/__main__.py#L1009-L1011

And also a timeout checker thread (which I assume is only the backup in case `resource` does not work): https://github.com/boxed/mutmut/blob/f63029b239fa7c8042b2a45c5191674df419cd31/mutmut/__main__.py#L864-L878

Windows does not support `resource` so maybe we need to only use `resource` on Linux and the timeout checker on Windows.

Also, we now reuse processes. So when killing a long-running process we need to ensure that we restart that worker, and that we report it as timeouted. The restarting *should* already work with the current implementation. But it is not reported as a result. So we likely need to track which process starts which task and then in `_remove_stopped_workers` we need to map the killed processes to the tasks they were executing, and report them as timed out (depending on the exit code).

Also note that I think mutmut reports the timeouted mutant as "killed" accidentally (I did not make a MRE yet, but it seems like it).

### Debugging different results

I've tested it on https://github.com/TOD-theses/traces_parser with a `# pragma: no mutate` on these lines (they would run into an infinite loop as we have no timeout yet): https://github.com/TOD-theses/traces_parser/blob/49d0b9cbc6fad490bba1f8a3136b24cd964336e1/traces_parser/parser/storage/memory.py#L35-L36

The main branch of mutmut outputted slightly different results than this windows-support branch of mutmut. I am not yet sure if this is due to some kind of race condition, or if it is deterministic. Maybe adding some additional logs would be useful to debug this.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,13 +4,13 @@ on: [push, pull_request]
 
 jobs:
   tests:
-    name: Test Python ${{ matrix.python-version }}
-    # Todo: Revert to ubuntu-latest when Python 3.7 support no longer needed
-    runs-on: ubuntu-22.04
+    name: Test ${{ matrix.os }} - ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.13", "3.12", "3.11", "3.10", "3.9"]
+        os: [ubuntu-latest, windows-latest]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/e2e_projects/config/pyproject.toml
+++ b/e2e_projects/config/pyproject.toml
@@ -23,7 +23,7 @@ dev = [
 ]
 
 [tool.mutmut]
-debug = true
+debug = false
 paths_to_mutate = [ "config_pkg/" ]
 do_not_mutate = [ "*ignore*" ]
 also_copy = [ "data" ]
--- a/e2e_projects/config/tests/test_main.py
+++ b/e2e_projects/config/tests/test_main.py
@@ -21,6 +21,6 @@ def test_max_stack_depth():
 def test_data_exists():
     path = (Path("data") / "data.json").resolve()
     assert path.exists()
-    with open(path) as f:
+    with open(path, encoding='utf-8') as f:
         data = json.load(f)
         assert data['comment'] == 'this should be copied to the mutants folder'
--- a/e2e_projects/my_lib/pyproject.toml
+++ b/e2e_projects/my_lib/pyproject.toml
@@ -17,4 +17,4 @@ dev = [
 ]
 
 [tool.mutmut]
-debug = true
+debug = false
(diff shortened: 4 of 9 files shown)


Otto-AA:
For the record: One / The main reason I discontinued working on this PR is, that re-using workers from a pool is more brittle. If I run mutant A in a process and this mutant breaks some global setup, then running mutant B in the process will produce wrong results. The `fork` method executes each mutant in their own sandbox, so if mutant A breaks some global setup, mutant B won't be affected by this.

But without worker pools, we'd likely need to reload all modules for each mutant (maybe even the python interpreter, though this could be possible with forkserver?), so the performance would be much worse.

Issue #368, opened 2025-03-03
tests_dir configuration is gone?


gurbentor:
Hi, 

I'm currently trying to migrate from mutmut V2 to V3, and I see that although it still appears in the README, the tests_dir part is no longer being parsed when mutmut is building the config and it just tries to run all tests un "tests/" dir.
It fails in my project setup since under "tests/" I have the dirs "unit/", "integration/", "load/", etc, and obviously I want mutmut to only use the unit tests. The other are not meant to run locally and fails immediately.
 
Am I missing something? Is there really no way to specify the tests dir we want to use?

Thanks

boxed:
Hmm. Yea, that's right. As mutmut 3 is now tightly integrated with pytest, that option didn't survive. I think it should be fairly easy to implement though. It looks like `PytestRunner.execute_pytest` could add the paths before the execution. A PR would be very welcome.


(Well.. load tests are obvious, but "integration" sounds like something you'd want to run with mutmut to me. That term normally means just things that might include DB or something like that.)

gurbentor:
Oh Ok, thanks for the response.
I hope I'll be able to get to it soon and create a PR.

PR #408, opened 2025-07-10
feat: add tests_dir to config

souradeep-das:
## Overview

closes https://github.com/boxed/mutmut/issues/368

## Changes

Adds `test_dir` to the config, allowing specific tests to be run

## Notes

The logic has been added to each high level method, instead of `PytestRunner.execute_pytest` thinking about fine-grained control. Do lmk if that needs to be changed


Otto-AA:
Awesome, thanks for the E2E test and the list fix!

Issue #419, opened 2025-08-10
No tests detected when testing mutants


CyberiaResurrection:
Mutmut version: 3.3.1
Python version: 3.9.21
OS: Fedora Linux 41

As it says on the tin - now I'm getting _no_ tests picked up for testing candidate mutants.  I get the same issue on mutmut 3.3.0.

Minimal(-ish) reproducing example: https://github.com/CyberiaResurrection/traveller_pyroute_mini/tree/64d5202df234598c6fd54d779ffe9c92c296a38a 

pytest.ini:
```
[pytest]
python_files = test*
python_classes = test*
pythonpath =
    ./

testpaths =
    Tests/SystemData/**.py
```

pyproject.toml:
```
[tool.coverage.run]
source = ["PyRoute/"]


[tool.mutmut]
paths_to_mutate = ["PyRoute/"]
tests_dir = ["Tests/"]
also_copy = ["Tests/", "pytest.ini"]
runner = "pytest -x"
max_stack_depth = 4
debug = true
```

Mutation run, after deleting `mutants/` folder + subfolders:
```
...
⠹ Running clean testspython -m pytest  -vv -x -q Tests/ --rootdir=.
========================================================================================================= test session starts =========================================================================================================
platform linux -- Python 3.9.21, pytest-8.4.1, pluggy-1.6.0 -- /home/alex/gitstuf/traveller_pyroute_mini/venv/bin/python
cachedir: .pytest_cache
Using --randomly-seed=4145466732
rootdir: /home/alex/gitstuf/traveller_pyroute_mini/mutants
configfile: pytest.ini
plugins: console-scripts-1.4.1, randomly-3.16.0, subtests-0.14.2
collected 55 items                                                                                                                                                                                                                    

Tests/SystemData/testStarList.py::testStarList::test_black_hole_no_space PASSED                                                                                                                                                 [  1%]
Tests/SystemData/testStarList.py::testStarList::test_neutron_star_no_space PASSED                                                                                                                                               [  3%]
Tests/SystemData/testStarList.py::testStarList::test_9_stars_none_trim PASSED                                                                                                                                                   [  5%]
Tests/SystemData/testStarList.py::testStarList::test_check_star_size_against_primary PASSED                                                                                                                                     [  7%]
Tests/SystemData/testStarList.py::testStarList::test_no_star_no_space PASSED                                                                                                                                                    [  9%]
[...]

PR #420, opened 2025-08-10
Tweak spin titles under debug

CyberiaResurrection:
I only really noticed this when reporting #419  , the bits in question being:

```
⠹ Running clean testspython -m pytest  -vv -x -q Tests/ --rootdir=.
```

and

```
⠸ Running forced fail testpython -m pytest  -vv -x -q Tests/ --rootdir=.
```

After more debug runs, I saw similar issues happened for `Generating mutants` , `Listing all tests` and `Running stats`.

This PR should be quite simple to understand - the current status quo prevails when `mutmut.config` is not set or `mutmut.config.debug` is False, but when it's True, append a newline to the spinner title in question so the next line of output starts _on the next line_, rather than appended to the end of the spinner title.

Checking `mutmut.config is None` first obviates the NoneType reference error that would otherwise occur - eg in mocked tests.

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -627,6 +627,8 @@ class CatchOutput:
     def __init__(self, callback=lambda s: None, spinner_title=None):
         self.strings = []
         self.spinner_title = spinner_title or ''
+        if mutmut.config is not None and mutmut.config.debug:
+            self.spinner_title += '\n'
 
         class StdOutRedirect(TextIOBase):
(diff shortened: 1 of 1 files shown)


boxed:
Couldn't this extra newline to the title be handled inside the CaptureOutput constructor? That would cut this change down to 2 lines in one place, no?

CyberiaResurrection:
That should be rebased now

PR #421, opened 2025-08-12
Bail out of stack traversal when hitting unittest files

CyberiaResurrection:
As at the status quo, stack traversal terminates when hitting a file with 'pytest' or 'hammett' in the path, but blithely carries on if a file has 'unittest' in the path, despite that being the unittest framework.

The simple fix to that is to also bail out if 'unittest' is in the current frame's filename.


boxed:
Thanks!

PR #422, opened 2025-08-13
feat: allow supplying options on cli tool

souradeep-das:
## Overview

Allow supplying paths-to-mutate (as comma separated values), tests-dir, and debug 'options' through the cli tool

Example:
```
mutmut --paths-to-mutate <path1,path2> --tests-dir <path> --debug run
```

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -689,7 +689,10 @@ class Config:
         return False
 
 
-def config_reader():
+def config_reader(cli_overrides=None):
+    if cli_overrides is None:
+        cli_overrides = {}
+
     path = Path('pyproject.toml')
     if path.exists():
         if sys.version_info >= (3, 11):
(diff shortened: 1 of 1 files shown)


souradeep-das:
Thanks for the feedback! Any particular context/reasons why this could be bad?

nicklafleur:
@souradeep-das I'm closing this PR due to inactivity. If this continues to be an issue for you feel free to re-open and we can discuss

Otto-AA:
> Sorry, not sure I follow. If I have a test function, test_really_really_long_running, how do I exclude it from the tests mutmut runs? It doesn't have a particularly deep stack compared to other tests, just a lot more iterations. In pytest, we use -m "not slow".

FYI, we added support for the config `pytest_add_cli_args_test_selection` which you could use as `pytest_add_cli_args_test_selection = ["-m", "not slow"]` in your pyproject.toml file. See the README for more information.

Issue #427, opened 2025-08-25
Trying to understand test timing


CyberiaResurrection:
I'm trying to understand why a rather non-trivial test gets recorded in `mutmut-results.json` with (presumably) sub-millisecond timing.

Actual test:
```
(venv) alex@localhost:~/gitstuf/traveller_pyroute$ time pytest -k test_landmarks_of_dagudashaag_and_zarushagar
========================================================================================================= test session starts =========================================================================================================
platform linux -- Python 3.9.21, pytest-8.4.1, pluggy-1.5.0
Using --randomly-seed=17648162
rootdir: /home/alex/gitstuf/traveller_pyroute
configfile: pytest.ini
testpaths: Tests/**.py, Tests/Hypothesis/**.py, Tests/Hypothesis/Inputs/**.py, Tests/Hypothesis/Position/**.py, Tests/Inputs/**.py, Tests/Mapping/**.py, Tests/Outputs/**.py, Tests/Pathfinding/**.py, Tests/Pathfinding/Schemas/**.py, Tests/Position/**.py, Tests/StatCalculation/**.py, Tests/SystemData/**.py, Tests/Utilities/**.py
plugins: anyio-4.9.0, console-scripts-1.4.1, typeguard-4.4.4, hypothesis-6.137.1, randomly-3.16.0, dash-2.18.2, subtests-0.14.2
collected 625 items / 624 deselected / 1 selected                                                                                                                                                                                     

Tests/Pathfinding/Schemas/testLandmarksExtremes.py .                                                                                                                                                                            [100%]

============================================================================================ 1 passed, 624 deselected, 7 warnings in 6.68s ============================================================================================
sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute

real    0m8.246s
user    0m7.234s
sys     0m0.678s

```

Much shorter test, to estimate pytest spinup time:
```
(venv) alex@localhost:~/gitstuf/traveller_pyroute$ time pytest -k testAccumulateSelf
========================================================================================================= test session starts =========================================================================================================
platform linux -- Python 3.9.21, pytest-8.4.1, pluggy-1.5.0
Using --randomly-seed=1924011032
rootdir: /home/alex/gitstuf/traveller_pyroute
configfile: pytest.ini
[...]

PR #429, opened 2025-08-26
Adjust duration_by_test definition to reflect new use

CyberiaResurrection:
As at the status quo, whichever call item ran last (usually 'teardown') had its time overwrite the test in question's entry in duration_by_test.

The simplest way I could think of to avoid that was to add each call item's duration to the total in duration_by_test. Changing duration_by_test to a defaultdict simplified the bookkeeping.

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -418,6 +418,10 @@ class PytestRunner(TestRunner):
     def run_stats(self, *, tests):
         class StatsCollector:
             # noinspection PyMethodMayBeStatic
+            def pytest_runtest_logstart(self, nodeid, location):
+                mutmut.duration_by_test[nodeid] = 0
+
+            # noinspection PyMethodMayBeStatic
             def pytest_runtest_teardown(self, item, nextitem):
                 unused(nextitem)
                 for function in mutmut._stats:


CyberiaResurrection:
@Otto-AA , @boxed , I think this addresses #427.

CyberiaResurrection:
@Otto-AA  - are you meaning between different test runs?

If so, isn't that (at least currently) mitigated by only passing the StatsCollector plugin to the stats-collection test run?
Or are you meaning unconditionally clearing `duration_by_test` at the start of stats collection?

Otto-AA:
I'm thinking about what happens when `duration_by_test` is already filled before `run_stats` is called. Previously we would have replaced existing test durations in `duration_by_test` (by using `=`), now we would add to it (by using `+=`)

This could happen if we call `run_stats` multiple times within a single command execution (we probably don't do this currently, maybe with the mutmut browse). Or by loading the stats from the json into `duration_by_test` and afterwards running `run_stats` (see https://github.com/boxed/mutmut/blob/5ab6b7785b6f2a54ebf6d79e4c3fa2ab3e619397/mutmut/__main__.py#L799). I think the latter case could also be fine because then we do not run it on all tests.

But it's easier to verify, and less likely to break in the future, if we set `mutmut.duration_by_test[item.nodeid] = 0` in a pytest_runtest_logstart hook in the `run_stats` method. So maybe your implementation is already working in all edge cases (I think so, but honestly don't know), but for me it seems easier to verify its correctness if we reset the duration to 0.

CyberiaResurrection:
@Otto-AA  , does that latest commit "Reset test duration at start of each test" do what you were thinking of?

Otto-AA:
Yes, looks good. Feel free to merge it if the stats still works with the reset (I can't test it locally).

You could also make the defaultdict a dict again if you want, it should work either way.

PR #430, opened 2025-08-27
Handle -9 PyTest return code

CyberiaResurrection:
This was more than a little weird - I was consistently getting a KeyError on line 589 of collect_stat, which maps return codes to a mutmut status.

More digging into the underlying boomage revealed the mutant that caused it, along with the tests that rumbled it.  For whatever reason, the process was getting killed _but not segfaulting_.  This manifested as a -9 return code, and thus kaboom in collect_stat.

To avoid generating another mutant status, and because "segfault" is the least-worst match, map the newly-problematic return code to "segfault".

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -84,6 +84,7 @@ status_by_exit_code = {
     152: 'timeout',  # SIGXCPU
     255: 'timeout',
     -11: 'segfault',
+    -9: 'segfault',
 }
 
 emoji_by_status = {


boxed:
Makes sense. Maybe the code should default to some "unknown kill state" for any return code it doesn't have a mapping for? Seems more robust.

Otto-AA:
Is "suspicious" (code 35) a specific case? If not, we could use this one as the default.

CyberiaResurrection:
@Otto-AA  - "suspicious" is a specific state, so that may not work.  I'll roll this PR up as is and then look at adding an "unknown" status in a seperate PR.

PR #432, opened 2025-08-28
Supply status for missing exit codes

CyberiaResurrection:
Following on from the work in "Handle -9 PyTest return code", discussion with @Otto-AA and @boxed revealed it would be desirable to handle missing return codes in full generality.

The simplest way I could think of to _do_ that, trying to eliminate side effects, is to convert status_by_exit_code to a defaultdict, with a default value of "suspicious" - one already defined, and handled, elsewhere in the codebase.

### DIFF
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -67,7 +67,7 @@ from mutmut.trampoline_templates import CLASS_NAME_SEPARATOR
 # TODO: hash of function. If hash changes, retest all mutants as mutant IDs are not stable
 
 
-status_by_exit_code = {
+status_by_exit_code = defaultdict(lambda: 'suspicious', {
     1: 'killed',
     3: 'killed',  # internal error in pytest means a kill
     -24: 'killed',
@@ -85,7 +85,7 @@ status_by_exit_code = {
     255: 'timeout',
     -11: 'segfault',
     -9: 'segfault',
-}
+})
 
 emoji_by_status = {
     'survived': '🙁',
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -1124,7 +1124,7 @@ def _run(mutant_names: Union[tuple, list], max_children: Union[None, int]):
             exit_code_by_key[mutant_name] = m.exit_code_by_key[mutant_name]
 
         for mutant_name, exit_code in sorted(exit_code_by_key.items()):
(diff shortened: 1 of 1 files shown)


Otto-AA:
You likely also have to make `exit_code_to_emoji` a defaultdict too

CyberiaResurrection:
> You likely also have to make `exit_code_to_emoji` a defaultdict too

If the defaultdict value for `status_by_exit_code` wasn't already _in_ `exit_code_to_emoji`, yes, definitely.  As I've gone with "suspicious", which is already in `exit_code_to_emoji` , I'm not (at a surface level) seeing why it should be a defaultdict.

CyberiaResurrection:
> Simple, but should work fine. You could test it by removing some other exit codes from the dict (eg the codes for killed) and then checking if it outputs suspicious instead of killed with the mutmut browse command.

Yes, it does - removing the exit code status for -9 results in that particular mutant being logged as "suspicious" instead of "segfault".

CyberiaResurrection:
> I think this line should use a direct dict lookup instead of the .get method (or simply `exit_code_to_emoji`):
> 
> https://github.com/boxed/mutmut/blob/229db0f7c64b76593d36bb799d426cef587126a2/mutmut/__main__.py#L1127
> 
> 
> I think with the .get() it won't use the defaultdict

According to https://docs.python.org/3/library/collections.html#collections.defaultdict , your reading of get() behaviour is correct.
On other hand, using the defaultdict lookup is a behaviour change above and beyond what I set out to do in this PR.
On third hand, the `?` method (not checked - ie, no tests run) _when receiving the result of a test run with a hitherto-unseen exit code_ could come across as misleading, as at least one test has actually run for the mutant in question.

Sod it, after writing that, I'll make the change you suggested.

Otto-AA:
We have following definition: 

```python
exit_code_to_emoji = {
    exit_code: emoji_by_status[status]
    for exit_code, status in status_by_exit_code.items()
}
```

This uses all items in `status_by_exit_code` to define `exit_code_to_emoji`. However, this does not copy the defaultdict behaviour. `exit_code_to_emoji` is still a normal dictionary and not a defaultdict, so it should fail when we access a non existing key.

CyberiaResurrection:
@boxed , @Otto-AA  , any objections to me merging this one?

boxed:
None. Go ahead.

Issue #435, opened 2025-09-08
Mutants identical to original method


CyberiaResurrection:
Mutmut version: 9f672be
Python version: 3.9.21
OS: Fedora Linux 41

Min reproducing example: https://github.com/CyberiaResurrection/traveller_pyroute_mini 

Unless I've seriously misunderstood (always a possibility), I'm of the understanding that mutation testing requires _changes_ to be made to program source code that can, at least in principle, be detected by a test.

With that understanding, I'm a little WTFed at the following segment of `mutants/PyRoute/Inputs/StarlineParser.py`:
```
    def xǁStarlineParserǁparse__mutmut_orig(self, text: str) -> tuple[Tree[Token], str]:
        text = re.sub(r'[\w\-]  [\w\-]', dashrepl, text)

        return self.parser.parse(text), text

    def xǁStarlineParserǁparse__mutmut_1(self, text: str) -> tuple[Tree[Token], str]:
        text = None

        return self.parser.parse(text), text

    def xǁStarlineParserǁparse__mutmut_2(self, text: str) -> tuple[Tree[Token], str]:
        text = re.sub(None, dashrepl, text)

        return self.parser.parse(text), text

<snip>


    def xǁStarlineParserǁparse__mutmut_9(self, text: str) -> tuple[Tree[Token], str]:
        text = re.sub(r'[\w\-]  [\w\-]', dashrepl, text)

        return self.parser.parse(text), text

    def xǁStarlineParserǁparse__mutmut_10(self, text: str) -> tuple[Tree[Token], str]:
        text = re.sub(r'[\w\-]  [\w\-]', dashrepl, text)

        return self.parser.parse(text), text
```
How can a mutant identical to the original (eg `xǁStarlineParserǁparse__mutmut_9`) contribute anything positive to mutation testing?

Where in (presumably) `file_mutation.py` would it be best to add a non-empty-diff check?

Just before line 121?
https://github.com/boxed/mutmut/blob/6c97a3e8a1fa5c709b52988d321e12ede160290f/mutmut/file_mutation.py#L121
(aka `MutationVisitor._create_mutations`)

_How_ would be the best way to do the non-empty-diff check?  Pull out the core of the existing `get_diff_for_mutant` function (after the `print` call) and what it calls to its own file, import & reference said new file in `__main__.py` and `file_mutation.py`, and add the call to `get_diff_for_mutant_core` to jsut before line 121 of `file_mutation.py` ?

[...]

PR #437, opened 2025-09-09
Fix unchanged string mutants

CyberiaResurrection:
As outlined in #435 , @Otto-AA stated that mutants being identical to the source code they're ostensibly mutating is a bug.

Again, I've tried to make the _simplest_, most tightly-focused, change I can think of to resolve the bug.


CyberiaResurrection:
Thanks for that @boxed.

PR #445, opened 2025-10-12
Add result descriptions in mutmut browse

Otto-AA:
Add result descriptions in mutmut browse.

- [x] add basic description for all results
- [x] add exit codes for all results (maybe helps with debugging / unexpected statuses)
- [x] add expected + actual time for timeouted mutants
- [ ] mention which test case killed a mutant
- [ ] show error for killed mutants (maybe only in mutmut show?)
- [x] somehow mention all relevant test cases, with some limit / scrolling / ...


Otto-AA:
I've added a new action "t" to view the tests for the selected mutant.

screenshot fullscreen

On a small screen, we will have a scroll option for the diff:

screenshot small

PR #460, opened 2026-01-21
Save mutation stats for use with CI/CD pipeline

mpownby:
This is a simple to change to save mutation stats after printing them so a CI/CD pipeline can optionally take action on the results.

Use case:
If mutation score is below a threshold, I want to block PR's from being merged and/or blocked a pipeline from advancing to the next stage.

Just testing the waters to see if you would be open to this change.  I read your contribution guidelines and didn't see anything that would make this an auto rejection.


boxed:
I think it makes more sense if this is a separate command. There is already saved data, so such a command could read that and write this simplified json for your use case.

Issue #465, opened 2026-02-01
Preserve original signature in trampoline


Otto-AA:
I think instead of copying signature and annotation to the trampoline function (see #464 ), we should construct a trampoline with an identical signature. This should be more robust than trying to copy.

For instance, for a method like this:

```python
def get_len(foo: int, bar: str):
    return ...
```

we currently create a trampoline like this:

```python
def get_len(*args, **kwargs):
    result = _mutmut_trampoline(x_get_len__mutmut_orig, x_get_len__mutmut_mutants, args, kwargs)
    return result
```

And I think we should move to something more like this:

```python
def get_len(foo: int, bar: str):
    result = _mutmut_trampoline(x_get_len__mutmut_orig, x_get_len__mutmut_mutants, args=[foo, bar], kwargs=...)
    return result
```

We probably should also preserve async keywords (see #454 ).

PR #468, opened 2026-02-14
Use type checking to detect invalid mutants

### LINKED ISSUE #467: Type hints/checking to reduce mutants
This issue is here to collect ideas, how mutmut could use type hints to prevent mutants that a type checker would catch.

For instance:

```python
def foo():
    x: int = 1
# should not be mutated to
def foo():
    x: int = None
```

#### 1. Using type hints while generating mutants

For simple cases like above, we could check the variable type hint, and if it's a basic type like `str`, `int`, ... then only allow these types. For more complex cases, I think we would need to use some type checker with an API and use it to understand which mutations are valid or not. I think this would get complex quickly and also the type checker we use would need to be compatible with the type checker of the user (as types are not 100% standardized).

#### 2. Generating mutants, then filtering out with type checker

I wonder if for (2), we could simply run a type checker and ignore mutants where the mutated method has a type error:

1. Generate mutants
2. Run type checker
3. Go through all lines with a type error
4. If the line is a mutated method, disable this mutant
5. For all remaining mutations, run pytest

The question I have here, is if we can ensure that mutations will only cause type errors in their own function. Could be possible, but not 100% sure.

At least our current implementation breaks it with pyright (see https://github.com/microsoft/pyright/blob/main/docs/type-inference.md#multi-assignment-type-inference):

```python
class Foo:
    def xǁFoo_foo__mutmut_orig(self):
        self.x = 'foo'

    def xǁFoo_foo__mutmut_1(self):
        self.x = None # mutated code
```

Originally, `Foo().x` is inferred to be of type `str`. With the mutated method in the same class, pyright will infer `Foo().x` to be the union `str | None`. This wlil break users of instance variables.

I think approach (2) is worth a try. If someone wants to work on it, let me know and I can give some pointers. Maybe I'll also start it when I have some free time :)

Otto-AA:
Fixes #467 

What works so far:
- [x] runs custom type checker on `./mutants/`
- [x] parses errors from JSON output (pyright, pyrefly)
- [x] maps error lines to mutants
- [x] disables mutants

Big TODOs:
- [ ] keep original typing of class properties (at least pyright types `self.x` as `str | None`, if there are multiple methods with `self.x = 'foo'` and `self.x = None`) (EDIT: see https://github.com/boxed/mutmut/pull/468#issuecomment-3910320357)
- [x] keep original typing of mutated methods (when doing `def foo(*args, **kwargs): return _trampoline...`, pyright cannot use the types of `foo` to infer types at places where we call `x = foo(a)`. This can break typing). Also see #465

I think for the class properties problem, we would need to define the mutated methods outside of the class and dynamically either add them or overwrite the original method. (EDIT: this won't work with `Self` types, `@classmethod`, etc.)

To keep the original types of mutated methods, using libcst to copy the signature should be fine.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -170,8 +170,16 @@ If you only want to mutate lines that are called (according to coverage.py), you
 Filter generated mutants with type checker
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-When your project is type checked, you can also use it to filter out invalid mutants.
+When your project is type checked using `mypy` or `pyrefly`, you can also use it to filter out invalid mutants.
 For instance, mutmut mutates `x: str = 'foo'` to `x: str = None` which can easily caught by type checkers.
+
+Using this filter can improve performance and reduce noise, however it can also hide a few relevant mutations:
+
+1. `x: str = None` may not be valid, but if your tests do not detect such a change it indicates that
+    the value of `x` is not properly tested (even if your type checker would catch this particular modification)
+2. In some edge cases with class properties (usually in the `__init__` method), the way `mypy` and `pyrefly` infer types does not work well
+    with the way mutmut mutates code. Some valid mutations like changing `self.x = 123` to `self.x = None` can
+    be filtered out, even though the may be valid.
 
 To enable this filtering, configure the `type_check_command` to output json results as follows:
 


Otto-AA:
I found one case where this filter can remove interesting mutations:

```python
class Foo:
    def __init__(self, a: int) -> None:
        self.a = a
```

mutates to:

```python
class Foo:
    def __init__(self, a: int) -> None:
        args = [a]# type: ignore
        kwargs = {}# type: ignore
        return _mutmut_trampoline(object.__getattribute__(self, 'xǁFooǁ__init____mutmut_orig'), object.__getattribute__(self, 'xǁStorageByteGroupǁ__init____mutmut_mutants'), args, kwargs, self)
    def xǁFooǁ__init____mutmut_orig(self, a: int) -> None:
        self.a = a
    def xǁFooǁ__init____mutmut_1(self, a: int) -> None:
        self.a = None # <- this is now a type error, because mypy/pyrefly already inferred 'int' on the first self.a = a definition
```

This should be a valid mutation, however it is marked as a type error. I can't think of a way to fix this, except for copying the whole class per mutation (similar to the `pyright`/`ty` problem described above).

And in general, I've noticed that even if a mutation has a type error, there are cases where it would show a lack of tests.

I've added these notes to the README. I think it can still be useful if there are too many survived mutants and you want to reduce the noise and/or improve performance. But it's not perfect, just a tradeoff.

PR #469, opened 2026-02-14
Preserve function signature

### LINKED ISSUE #454: Async functions get synchronous trampolines causing "coroutine was never awaited" errors
## Description

When using mutmut 3.4.0 with async functions (`async def`), the generated trampoline wrapper is synchronous (`def`), which causes the coroutine to not be awaited.

## Environment

- mutmut version: 3.4.0
- Python version: 3.12.3
- OS: Linux (WSL2)

## Steps to Reproduce

1. Have a codebase with async functions like:

```python
async def get_current_user(
    credentials: Annotated[HTTPAuthorizationCredentials, Depends(security)],
    db: Annotated[AsyncSession, Depends(get_db)],
) -> User:
    """Get current authenticated user from JWT token."""
    # ... async code
    user = await service.get_user_by_id(uuid.UUID(user_id))
    return user
```

2. Run `mutmut run`

3. Observe the generated trampoline in `mutants/` directory

## Expected Behavior

The trampoline for async functions should be:

```python
async def get_current_user(*args, **kwargs):
    result = await _mutmut_trampoline(x_get_current_user__mutmut_orig, x_get_current_user__mutmut_mutants, args, kwargs)
    return result
```

## Actual Behavior

The generated trampoline is synchronous:

```python
def get_current_user(*args, **kwargs):
    result = _mutmut_trampoline(x_get_current_user__mutmut_orig, x_get_current_user__mutmut_mutants, args, kwargs)
    return result 
```

This causes the error:
```
RuntimeWarning: coroutine 'x_get_current_user__mutmut_orig' was never awaited
```

And tests fail because the coroutine is returned instead of being awaited.

## Root Cause

In `trampoline_templates.py`, the `build_trampoline` function always generates a synchronous `def`:

```python
def {orig_name}({'self, ' if class_name is not None else ''}*args, **kwargs):
    result = {trampoline_name}(...)
    return result 
```

It should detect if the original function is async and generate:

```python
async def {orig_name}({'self, ' if class_name is not None else ''}*args, **kwargs):
    result = await {trampoline_name}(...)
    return result 
```

## Related

- #406 - Fixed async generators, but not regular async functions
- #407 - PR that added async generator support

## Workaround

Currently excluding files with async functions using `do_not_mutate` in `pyproject.toml`:

```toml
[tool.mutmut]
do_not_mutate = [
    "**/dependencies.py",
    "**/routes.py", 
    "**/service.py",
    "**/database.py",
]
```

This significantly limits mutation testing coverage for async Python codebases (FastAPI, asyncio, etc.).

Otto-AA:
Closes #454 and #465 .

@boxed In case you want to review, this is a slightly bigger change of trampoline internals. This PR adds some complexity at the trampoline wrapper (handling various possible parameter formats, async declarations and async generators), but I think it's worth it so we can preserve the original function signature. Also, don't need to manually copy the signature with `trampoline.__signature__ = inspect.signature(original_method)` anymore.

With this change, `asyncio.iscoroutinefunction(async_func)` also returns true after we mutated the function. This is done by using the same signature as it was originally, and adding some logic to pass our arguments to the trampoline. 

For instance:

```python
async def foo(a: str, b, *args, **kwargs) -> dict[str, int]:
    # ...
```

Will be mutated to:

```python
async def foo(a: str, b, *args, **kwargs) -> dict[str, int]:
    args = [a, b, *args]
    kwargs = {**kwargs}
    return await _mutmut_trampoline(x_foo__mutmut_orig, x_foo__mutmut_mutants, args, kwargs, None)
```

In the commit 944d43fa2bf198ac1156804984838784e9e302fb, I've also added `inline-snapshot` (+ruff formatting for the snapshots) to make it easier to work with regression snapshots. e.g. the `tests/test_mutation regression.py` are useful to have and with `inline-snapshot` it's easier to update them whenever some implementation detail changes.

The e2e tests and a sample repository I tried were not affected by this change, so I don't think it breaks existing functionality.

### DIFF
--- a/e2e_projects/my_lib/src/my_lib/__init__.py
+++ b/e2e_projects/my_lib/src/my_lib/__init__.py
@@ -104,9 +104,9 @@ def some_func(a, b: str = "111", c: Callable[[str], int] | None = None) -> int |
         return c(b)
     return None
 
-def func_with_star_clone(a, *, b, **kwargs): pass  # pragma: no mutate
-def func_with_star(a, *, b, **kwargs):
-    return a + b + len(kwargs)
+def func_with_star_clone(a, /, b, *, c, **kwargs): pass  # pragma: no mutate
+def func_with_star(a, /, b, *, c, **kwargs):
+    return a + b + c + len(kwargs)
 
 def func_with_arbitrary_args_clone(*args, **kwargs): pass  # pragma: no mutate
 def func_with_arbitrary_args(*args, **kwargs):
--- a/e2e_projects/my_lib/tests/test_my_lib.py
+++ b/e2e_projects/my_lib/tests/test_my_lib.py
@@ -61,7 +61,7 @@ def test_that_signatures_are_preserved():
 
 def test_signature_functions_are_callable():
     assert some_func(True, c=lambda s: int(s), b="222") == 222
-    assert func_with_star(1, b=2, x='x', y='y', z='z') == 6
+    assert func_with_star(1, b=2, c=3, x='x', y='y', z='z') == 9
     assert func_with_arbitrary_args('a', 'b', foo=123, bar=456) == 4
 
 def test_signature_is_coroutine():
--- a/src/mutmut/file_mutation.py
+++ b/src/mutmut/file_mutation.py
@@ -258,10 +258,10 @@ def function_trampoline_arrangement(function: cst.FunctionDef, mutants: Iterable
 
 def create_trampoline_wrapper(function: cst.FunctionDef, mangled_name: str, class_name: str | None) -> cst.FunctionDef:
     args: list[cst.Element | cst.StarredElement] = []
-    for param in function.params.params:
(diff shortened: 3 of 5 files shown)


boxed:
Nice!

Peter-Lavigne:
I think this PR introduced a regression for default-value mutations. With the old *args, **kwargs trampoline, a mutation to a default parameter was observable because the mutant function resolved its own default. With the new signature-preserving trampoline, the wrapper resolves the default before dispatch, making the mutation unreachable.

For example, given this source:

```
def check(fix: bool = False) -> None:
    ...
```

The generated trampoline is:

```
def check(fix: bool = False) -> None:
    args = [fix]
    kwargs = {}
    return _mutmut_trampoline(x_check__mutmut_orig, x_check__mutmut_mutants, args, kwargs, None)
```

I'm new to mutation testing and to this repo, so let me know if I'm missing something.

Issue #414, opened 2025-07-19
Clearly document what `paths_to_mutate` does (or rename it)


Otto-AA:
In several issues `paths_to_mutate` was used to select specific files that should be mutated (For instance, see #370). Given the name, this is a natural interpretation of what it does, and the docs do not clearly say otherwise. However, mutmut uses it as paths to the source code (/packages / modules; not sure about the terminology), and it needs to contain all the code necessary to run the test suite.

I would consider to rename it (e.g. to `source_paths`) and make `paths_to_mutate` a deprecated alias for it, showing a warning that suggests switching to `source_paths`.

Also, apparently people want to configure which files are mutated in the source code. Similar to the `do_not_mutate` config, we could have a `only_mutate` option, taking a list of paths that we mutate. Something similar is possible with the CLI positional arg (e.g. `mutmut run "narwhals._arrow.series*"`), but this only changes which mutants get tested, not which files are mutated on the disk.

Issue #367, opened 2025-03-01
Implement shadowing


boxed:
Deletion of an entire line should shadow all other mutants of that line. Same goes for assigning mutated to just assign to None if the original expression is a function call or similar complex statement.

Otto-AA:
I think there are two parts here we could implement:
1. use shadowing to hide (collapse/...) mutants in the `mutmut browse` output
2. use shadowing to prioritize and skip mutation runs (first run the mutation that deletes a whole line; only test other mutations if the line deletion mutant got killed)

For the start, I think (1) would be good enough. Skipping mutation runs could be added on top of it afterwards.

One idea would be to annotate each mutation with a range in the source code (from line 123 col 10 until line 124 col 30). Then maybe we can use `range(mutant A) includes range(mutant B)` as a metric for `mutant A shadows mutant B`?

Open to other ideas, and maybe there's inspiration in other tools or papers :)

Issue #47, opened 2018-11-26
(Optionally) skip mutations of constants


blueyed:
A lot of surviving mutants in a project are due to changing constants, e.g. `FOO = 1`, where `FOO` itself is used in tests then also.
Killing those would need a test where the value itself is used, which does not make much sense, does it?

I think it would be good if those mutations could be skipped (e.g. by not mutating anything that matches `[A-Z_]+` maybe?

It would be good if those could be deselected either by name of the mutation (which would need to be something like `mutate-constant-X`), or by specifying a pattern of identifiers not to mutate.
(it is not possible to deselect mutations currently, is it?)

boxed:
This seems like mostly the same as what we discussed in https://github.com/boxed/mutmut/issues/28 but for different reasons obviously. Maybe one should be able to just plug in an arbitrary python function that gets called to ask for exclusion?

boxed:
Actually, thinking more about this it would be interesting if you could explain more about your situation. In my experience mutating constants is one of the more useful things for finding lacking tests or even outright bugs. I don't understand your description.

blueyed:
If you have `EXIT_OK = 0` and then check if your program exits with `EXIT_OK` in some case, the test will also pass with `EXIT_OK = 1` (because the constant is used in the test also).
You would need a test that `EXIT_OK == 0`, or use the number itself in the test.  But in this case the constant more or less defines the behavior.

boxed:
Ah, I see. Well yea, in that case the current system to deal with it is `# pragma: no mutate`. What do you think about more advanced exclusions like having a callback you can specify in the config?

boxed:
Yea, maybe a global and built in whitelist would be a good idea. I have another candidate for that:

```python
__import__('pkg_resources').declare_namespace(__name__)
```

we always have to put a pragma no mutate on those.

boxed:
I've released a new version of mutmut that ignored a bunch of these by default. It's also got an overhauled cache system that is way better. Give it a try!

lundybernard:
A related issue that I'm dealing with: I want to skip mutations on log messages.
I normally avoid making assertions about log messages, as it feels more like locking down the code than a behavioral test.

ex:
```
 log.debug(
     'whatever message you like, maybe its long and'
-     'broken up over multiple lines'
+    'XXbroken up over multiple lines'
 )
```
`# pragma: no mutate` isn't a good option at a cost of 19 characters per line

blueyed:
[...]

PR #471, opened 2026-02-25
Preserve cached mutation results on rerun of unchanged source

percy-raskova:
## Summary

`create_mutants_for_file()` correctly detects unchanged source files (`source_mtime < mutant_mtime`) and skips regenerating mutations, but also resets all `exit_code_by_key` values to `None`. This causes every subsequent `mutmut run` to re-test all mutants from scratch, defeating the incremental caching introduced in bf3a2e81.

**Root cause:** Lines 296-301 in `__main__.py` load the `.meta` file and set every exit code to `None` before returning `unmodified=True`. The skip logic at line 1273 (`if result is not None: continue`) then sees all-`None` results and re-tests everything.

**Fix:** Remove the reset loop so cached results survive across runs.

## Changes

- **Bug fix:** Remove the exit code reset in the `source_mtime < mutant_mtime` branch of `create_mutants_for_file()`
- **Test:** Add e2e test that injects sentinel exit codes into `.meta` files and verifies they survive a second run on unchanged source
- **Lint:** Fix pre-existing ruff errors (unused imports `Iterable`/`libcst.matchers`, unused variable `tests_dir`, f-strings without placeholders)

## Test plan

- [x] New e2e test `test_rerun_preserves_cached_results` — RED without fix, GREEN with fix
- [x] All 179 existing tests pass
- [x] All 6 e2e tests pass (including existing snapshot tests)
- [x] `ruff check` clean on changed files

### DIFF
--- a/src/mutmut/__main__.py
+++ b/src/mutmut/__main__.py
@@ -1,4 +1,3 @@
-from typing import Iterable
 from mutmut.type_checking import TypeCheckingError
 from mutmut.type_checking import run_type_checker
 from typing import Any
@@ -55,7 +54,6 @@ import warnings
 
 import click
 import libcst as cst
-import libcst.matchers as m
 from rich.text import Text
 from setproctitle import setproctitle
 
@@ -137,7 +135,7 @@ def guess_paths_to_mutate():
 
 
 def record_trampoline_hit(name):
-    assert not name.startswith('src.'), f'Failed trampoline hit. Module name starts with `src.`, which is invalid'
+    assert not name.startswith('src.'), 'Failed trampoline hit. Module name starts with `src.`, which is invalid'
     if mutmut.config.max_stack_depth != -1:
         f = inspect.currentframe()
         c = mutmut.config.max_stack_depth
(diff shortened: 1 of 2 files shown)


Otto-AA:
Hey @percy-raskova , I did not intend to say that your wokr is not useful, sorry if it came across this way. My main point was that I think it should be opt-in (as this type of incremental testing can skip relevant mutants) and that I frankly haven't thought through this problem yet and don't know what's a reasonable approach. I think now it makes sense to wait for how @nicklafleur approached it and decide which way we want to go. It's also interesting to compare to other tools for inspiration (e.g. https://stryker-mutator.io/docs/stryker-js/incremental/ and https://infection.github.io/guide/command-line-options.html#git-diff-filter).

If you want to work on something else in the meantime, feel free to take a look at the issue tracker and ask for pointers or explanations on how the code base works. For instance, #414 would be a rather small change (renaming `paths_to_mutate` because the name can be confusing), or #47 if you want to go deeper into how we use libCST to parse code. Or as @boxed explained, you could also take a look at shadowing in #367.

If you have any questions, just ask :) (but it can take some days until I reply)

Otto-AA:
> Even though I'm the original author of mutmut I seem to be more of an vision/idea man at this point, with Otto-AA doing the actual work on the system and he has a much better living understanding of the tradeoffs and implementation than me now :P

I have only used mutmut on side projects and there's not much user feedback, so my UX tradeoffs are probably not the best :upside_down_face: I also feel like string mutations got a bit much, we probably want to enable/disable some of them via settings and per default. But I enjoy thinking about the problems and how things could be improved, and want to get it working on some production code bases at some point.

nicklafleur:
@percy-raskova sorry about the delay, but https://github.com/boxed/mutmut/pull/509 contains my take on result caching. Would love your review when you have the chance.

nicklafleur:
@percy-raskova the ast-based caching has been merged to main, sorry about the long wait, life happened. I believe that this issue has been resolved with the current implementation, so I will close this PR but if you still find issues or believe that the solution can use more work, please feel free to open a new one.

PR #476, opened 2026-03-01
feat: Support Enum, @staticmethod, @classmethod mutations

nicklafleur:
Mix of boy-scouting and new features along with docker improvements for non-linux systems. 

Summary
- Refactor Config to a singleton pattern — Extract configuration into a dedicated Config class with get()/reset()/ensure_loaded() API, replacing the global mutmut.config (kept behind a deprecation warning). 
- Add type annotations throughout and a safe_setproctitle wrapper for macOS Python 3.14+ fork crashes.
- Fix timeout checker bug — Replace the pid-to-mutant lookup (which used a stale loop variable, causing hangs in parallel runs) with a min-heap based design that tracks (deadline, pid) tuples directly.
- Add enum and static/classmethod mutation support — Introduce an external injection pattern for enum class mutation (avoiding metaclass conflicts), allow @staticmethod/@classmethod through the decorator filter, and add pragma: no mutate class/pragma: no mutate function pragmas.
- Refactor mutation internals — Extract pragma_handling.py, enum_mutation.py, trampoline_templates.py, format_utils.py, and timeout.py into focused modules.
- Improve Docker test image — Copy pyproject.toml/uv.lock before uv sync so source changes don't re-download dependencies.
- /scripts/run_tests.sh now accepts multiple python versions for comprehensive testing workflows.

### DIFF
--- a/src/mutmut/mutation/file_mutation.py
+++ b/src/mutmut/mutation/file_mutation.py
@@ -614,7 +614,23 @@ class StringifyAnnotations(cst.CSTTransformer):
         if isinstance(updated_node.annotation, (cst.SimpleString, cst.ConcatenatedString, cst.FormattedString)):
             return updated_node
         source = self._empty_module.code_for_node(updated_node.annotation)
-        return updated_node.with_changes(annotation=cst.SimpleString(f'"{source}"'))
+        return updated_node.with_changes(annotation=cst.SimpleString(self._format_source(source)))
+
+    def _format_source(self, source: str):
+        """Format the source annotation accounting for types with quotes inside the outer annotation
+
+        EX: List[ "NewClass" ] -> "List[ \"NewClass\" ]"
+
+        Attempts to default to more readable approaches but falls-safe to a globally
+        compatible approach regardless of quote style (including the seldom-used in this context triple quote).
+        """
+        if '"' not in source:
+            return f'"{source}"'
+        elif "'" not in source:
+            return f"'{source}'"
+        else:
+            escaped = source.replace('"', '\\"')
+            return f'"{escaped}"'
 
 
 def _is_generator(function: cst.FunctionDef) -> bool:
--- a/tests/mutation/test_mutation_runtime.py
+++ b/tests/mutation/test_mutation_runtime.py
@@ -48,6 +48,55 @@ class Color(Enum):
             del os.environ["MUTANT_UNDER_TEST"]
 
 
+def test_type_annotation_runtime_execution():
+    """Test that forward and quoted annotations can be executed
+
+    We don't really care about any mutations, we just care if the annotations are processed properly.
+    """
+    source = """
+from enum import Enum
+from typing import List, Literal
+
+class Color(Enum):
+    RED = 1
+    GREEN = 2
+    BLUE = 3
+
+    def describe(self) -> Literal[ "red" | 'green' | \"""blue\""" ]: # worst case type annotation
+        return self.name.lower()
+
+    @classmethod
+    def from_name(cls, name: str) -> Color:
+        name = name.upper()
+        vals = {
+            Color.RED.name: Color.RED,
+            Color.GREEN.name: Color.GREEN,
+            Color.BLUE.name: Color.BLUE
+        }
+        return vals[name]
+""".strip()
+
+    mutated_code, mutant_names = mutate_file_contents("test.py", source)
+    assert len(mutant_names) > 0, "Should have at least one mutant"
+
+    old_env = os.environ.get("MUTANT_UNDER_TEST")
+    try:
+        os.environ["MUTANT_UNDER_TEST"] = "none"
+        namespace = {"__name__": "test_module"}
+        exec(mutated_code, namespace)
+        Color = namespace["Color"]
+
+        assert Color.RED.value == 1
+        assert Color.RED.describe() == "red"
+
+        assert Color.from_name("green") == Color.GREEN
+    finally:
+        if old_env is not None:
+            os.environ["MUTANT_UNDER_TEST"] = old_env
+        elif "MUTANT_UNDER_TEST" in os.environ:
+            del os.environ["MUTANT_UNDER_TEST"]
+
+
 def test_regular_class_staticmethod_runtime():
     """Test that staticmethod mutation in regular classes works at runtime."""
     source = """


nicklafleur:
> > Fix timeout checker bug — Replace the pid-to-mutant lookup (which used a stale loop variable, causing hangs in parallel runs) with a min-heap based design that tracks (deadline, pid) tuples directly.
> 
> Can you explain which bug this solves? How did the pid lookup become stale (did someone else modify the dict in the same time?) and how does this cause a hang?

Hey sorry I should have included more details in the PR, here's what I added in the commit [40eb31d](https://github.com/boxed/mutmut/pull/476/commits/40eb31d448747225bd3128858a4d3d8bdab1c640).
```text

The old timeout_checker used the loop variable `mutant_name`
from the outer mutants iteration when looking up estimated_time_of_tests,
instead of looking up the mutant name associated with each PID, causing
some tests to hang and never be killed due to incorrect timeout
calculations.

This was easy to miss because the bug only manifested when multiple mutants
were being tested in parallel and the timing was just right for the wrong
mutant's timeout to be checked against another mutant's PID (in my case,
it only happened specifically when running the e2e tests in vscode
after my last set of changes 🤷)

Instead of trying to patch this specific bug (and probably introducing
one), I decided to go with a simpler design that avoids trying to find
the right pid at all, instead using a min-heap to track (timeout, pid)
tuples.
```

src/mutmut/utils/safe_setproctitle.py
```diff
@@ -0,0 +1,26 @@
...
+
+This module provides a safe_setproctitle() function that:
+- Works normally on Linux
+- Is a no-op on macOS to avoid crashes after fork()
+"""
+
+import platform
+import sys
+
+from setproctitle import setproctitle as _setproctitle
```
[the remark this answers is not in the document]
nicklafleur: handled in [a8cf8dd](https://github.com/boxed/mutmut/pull/476/commits/a8cf8dde8787799e9c8d6780733170f6459c1521)

README.rst
```diff
@@ -226,6 +270,92 @@ whitelist lines are:
...
+
+**Disabling Enum Mutation**
+
+If you prefer to skip enum mutation entirely, you can disable it in your config:
+
+.. code-block:: toml
+
+    # pyproject.toml
+    [tool.mutmut]
+    mutate_enums = false
```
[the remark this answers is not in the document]
Otto-AA: I understand the rationale, but for simplicity (both of our implementation and also number of configs+documentation) I'd remove this config. If we do this for all new features we will end up with very many options (for disabling enums, staticmethods, ternary operators, etc.). I think if we expect the mutation to work in all setups and be useful we don't need extra individual configs.

At some point, we'll have #47 and probably a config to enable/disable specific mutations, but with a common interface.

src/mutmut/mutation/file_mutation.py
```diff
@@ -354,6 +516,53 @@ def _get_local_name(func_name: str) -> cst.BaseExpression:
...
+        if not method_mutants:
+            new_body.append(method)
+            continue
+
+        method_type = get_method_type(method)
+        if method_type is None:
+            new_body.append(method)
+            continue
+
+        ext_nodes, assignment, method_mutant_names = _external_method_injection(
```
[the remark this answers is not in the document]
Otto-AA: Thanks, it would be nice to have a more generalized version. But it's not a requirement for this PR, so no pressure if you don't really want to work on this or do it sometime later :)

I would guess that defining the mutants dict outside the class + methods inside the class should work for all cases. For "normal" methods this should work fine. For staticmethod we'd need to make sure we don't pass `self` as an argument. For classmethod we'd need to pass `cls` instead. For enums this should also work fine without any changes I think. Open to other suggestions though

README.rst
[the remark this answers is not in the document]
nicklafleur: makes sense to me!

src/mutmut/mutation/trampoline_templates.py
```diff
@@ -0,0 +1,183 @@
+from mutmut.mutation.mutators import MethodType
+
+CLASS_NAME_SEPARATOR = "ǁ"
+
+GENERATED_MARKER = "# mutmut: generated"
```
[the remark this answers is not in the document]
nicklafleur: That makes sense, I had considered something like that initially, but was thinking about trying to make it a bit more explicity _why_ we are ignoring it. I think I'll move to something like `#type: ignore  # mutmut generated` so that we adhere to PEP while also getting that additional signal.

src/mutmut/mutation/file_mutation.py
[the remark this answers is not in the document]
nicklafleur: yeah I was testing out what this would look like and it results in some really clean outputs. It becomes extremely clear which methods are being mutated vs not when the mutated code is moved outside. Since that does look to be a bit more involved of a PR at this point I'll submit it as a follow up later so that these changes can get merged without too much drift and to separate the cleanup a bit more explicitly.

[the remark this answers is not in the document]
Otto-AA: Okay, then let's skip this part for this PR.
A consideration for the future: moving mutated methods outside of the class likely does not work well with type checking (accessing `self`, `cls` or using `typing.Self` when not inside of the class body likely confuses type checkers). I would guess keeping them inside is easier.

src/mutmut/mutation/file_mutation.py
```diff
@@ -210,6 +283,10 @@ def combine_mutations_to_source(module: cst.Module, mutations: Sequence[Mutation
     for statement in remaining_statements:
         if isinstance(statement, cst.FunctionDef):
             func = statement
+            # Skip entire function if it has pragma: no mutate function
+            if func.name.value in ignored_functions:
+                result.append(func)
+                continue
```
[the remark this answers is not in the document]
nicklafleur: good catch!

nicklafleur:
@Otto-AA sorry about the long turnaround on this, ended up finding a bug with how the external mutation handled forward reference type-hints that needed to be fixed before pushing. Also spent a little time doing type-fu to get the trampolines to properly display types in IDEs, admittedly didn't look too much into how this could be used to augment the type-based mutant invalidation as this PR is already big enough.

I tried to separate the moved stuff from `__main__.py` to the other modules but the changes ended up being a bit too intermingled to do so, I'll be sure to be more conscious of that with future PRs.

Otto-AA:
Thanks for the updates! I don't think I'll have time to review this weekend, but hopefully in the course of the next week.

nicklafleur:
> Thanks for the updates! I don't think I'll have time to review this weekend, but hopefully in the course of the next week.

no problem, I'm still polishing things a bit today anyways so I'll happily take the time

tests/threading/test_timeout.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class TestLazyStart:
+    """Tests for lazy-start behavior of the timeout checker."""
+
+    def test_checker_starts_on_first_registration(self) -> None:
+        """Verify the checker thread starts automatically on first register_timeout call."""
+        timeout._checker_started = False
+
+        timeout.register_timeout(pid=12345, timeout_s=10.0)
```
Otto-AA: Are we killing a random process here? 12345 would be in the realm of pids 👀 
But I guess pytest exits before the Thread tries to kill the process with pid 12345 after 10s. I'll probably refactor this test a bit (maybe increase pid numbers and let register_timeout return the Thread? Not sure yet)

src/mutmut/mutation/file_mutation.py
```diff
@@ -243,6 +326,59 @@ def combine_mutations_to_source(module: cst.Module, mutations: Sequence[Mutation
...
+    :param method_type: MethodType.STATICMETHOD, MethodType.CLASSMETHOD, or MethodType.INSTANCE
+    :return: A tuple of (external_nodes, class_body_assignment, mutant_names)
+    """
+    external_nodes: list[MODULE_STATEMENT] = []
+    mutant_names: list[str] = []
+    method_name = method.name.value
+    prefix = f"_{class_name}_{method_name}"
+    mangled_name = mangle_function_name(name=method_name, class_name=class_name) + "__mutmut"
+
+    stringify = StringifyAnnotations()
```
Otto-AA, line 355: Could we also put the mutated methods + dictionary after the class definition? If yes, I think I'll change that so we don't need to stringify the annotations.

src/mutmut/mutation/file_mutation.py
```diff
@@ -243,6 +326,59 @@ def combine_mutations_to_source(module: cst.Module, mutations: Sequence[Mutation
...
+
+    orig_func = method.with_changes(name=cst.Name(f"{prefix}_orig"), decorators=[])
+    orig_func = cast(cst.FunctionDef, orig_func.visit(stringify))
+    external_nodes.append(orig_func)
+
+    for i, mutant in enumerate(mutants):
+        mutant_func_name = f"{prefix}_mutant_{i + 1}"
+        full_mutant_name = f"{mangled_name}_{i + 1}"
```
Otto-AA, line 363: Is there a reason we use a different naming scheme for these external mutated methods? If yes, we need to modify the `is_mutated_method_name` method s.t. type checking can match type errors in mutated methods to the mutant names. Could be E2E tested by adding a `@staticmethod` to the type_checking tests.

tests/threading/test_timeout.py
nicklafleur: ....uhhhh maybe (just kidding you're absolutely right, I should be using a patch for this, good catch)

src/mutmut/mutation/file_mutation.py
nicklafleur: I did try that because I found it visually cleaner too but unfortunately you can't "hoist" function definitions in the same way that you can have forward references for types so this seems to be the only way to actually do this though I'm totally open to other ideas.

src/mutmut/mutation/file_mutation.py
nicklafleur: probably just moved a little too fast on this one and didn't consider the side-effects. I'll tweak this to make it work with the type checking mechanisms tomorrow/Saturday.

tests/threading/test_timeout.py
nicklafleur: so I changed the fixture for those tests to clear the heap before and after tests and have really long timeouts, which should ensure that we never actually have a pid in the heap to kill when the thread wakes up

src/mutmut/mutation/file_mutation.py
Otto-AA: Ah I see, makes sense. If we can use a trampoline function inside the class body, the trampoline could forward reference to the mutated dicts/methods (like we do for normal class methods).

But that's not so important, I was just curious. I'd say that's a small cleanup potential for the future.

tests/threading/test_timeout.py
Otto-AA: Thanks for the update!

tests/mutation/test_pragma_handling.py
```diff
@@ -0,0 +1,247 @@
...
+
+def parse_pragmas(filename: str, source: str) -> tuple[set[int], set[int]]:
+    module = cst.parse_module(source)
+    wrapper = MetadataWrapper(module)
+    visitor = PragmaVisitor(filename)
+    wrapper.visit(visitor)
+    return visitor.no_mutate_lines, visitor.ignore_node_lines
+
+
+class TestParsePragmaLines:
```
dgrant, line 19: There's like 3 or 4 different stylings of `source` going on here.

This one looks like the easiest to read except for the `\n` characters:

```
        source = (
            "class Skipped:  # pragma: no mutate block\n"
            "    def method(self):\n"
            "        return 1 + 1\n"
            "\n"
            "def skipped_func():  # pragma: no mutate block\n"
            "    return 2 + 2\n"
            "\n"
            "def mutated():\n"
            "    return 3 + 3  # pragma: no mutate\n"
        )
```

Maybe this is better:

```
        source = """
            class Skipped:  # pragma: no mutate block
                def method(self):
                    return 1 + 1

            def skipped_func():  # pragma: no mutate block
                return 2 + 2

            def mutated():
                return 3 + 3  # pragma: no mutate
        """
```

tests/mutation/test_pragma_handling.py
```diff
@@ -0,0 +1,247 @@
...
+        no_mutate, ignore_node_lines = parse_pragmas("test.py", source)
+        assert no_mutate == set()
+        assert ignore_node_lines == set()
+
+
+class TestBlockPragma:
+    """Tests for # pragma: no mutate block."""
+
+    def test_own_line(self):
+        source = "if condition:\n    # pragma: no mutate block\n    x = 1\n    y = 2\nz = 3\n"
```
dgrant:
```suggestion
        source = """
        if condition:
            # pragma: no mutate block
            x = 1
            y = 2
            z = 3
        """
```

Same for all these below

tests/mutation/test_pragma_handling.py
nicklafleur: yeah good call, ended up manipulating stuff for testing a bit so things got weird but should have standardized it afterwards

nicklafleur: handled in https://github.com/boxed/mutmut/pull/476/commits/f54e066653d62712fa4dfb5e227312ae56fd2705

nicklafleur:
@Otto-AA The pragma rewrite has cause this PR to balloon considerably beyond what I would generally be comfortable submitting, if you would like me to split the enum/classmethod/staticmethod support from the pragma features just let me know and I'll do so to make it reviewable.

src/mutmut/mutation/file_mutation.py
```diff
@@ -0,0 +1,731 @@
...
+    This allows for mutations to be placed anywhere in the file, improving the resilience of
+    the mutation process without breaking type checking."""
+
+    _empty_module = cst.parse_module("")
+
+    def leave_Annotation(self, original_node: cst.Annotation, updated_node: cst.Annotation) -> cst.Annotation:
+        if isinstance(updated_node.annotation, (cst.SimpleString, cst.ConcatenatedString, cst.FormattedString)):
+            return updated_node
+        source = self._empty_module.code_for_node(updated_node.annotation)
+        return updated_node.with_changes(annotation=cst.SimpleString(f'"{source}"'))
```
nkakouros: This breaks annotations that contain double quotes, for example:

```
Literal["one", 'two']
```

nicklafleur: Good catch and thanks for the additional test case!

nicklafleur: fixed in [018e2ad](https://github.com/boxed/mutmut/pull/476/commits/018e2ad241888e11610b873b3e24fde532436a42)

nkakouros: Thank you for the fast response and fix! 🚀

src/mutmut/mutation/file_mutation.py
```diff
@@ -614,7 +614,23 @@ def leave_Annotation(self, original_node: cst.Annotation, updated_node: cst.Anno
         if isinstance(updated_node.annotation, (cst.SimpleString, cst.ConcatenatedString, cst.FormattedString)):
             return updated_node
         source = self._empty_module.code_for_node(updated_node.annotation)
-        return updated_node.with_changes(annotation=cst.SimpleString(f'"{source}"'))
+        return updated_node.with_changes(annotation=cst.SimpleString(self._format_source(source)))
+
+    def _format_source(self, source: str):
+        """Format the source annotation accounting for types with quotes inside the outer annotation
+
+        EX: List[ "NewClass" ] -> "List[ \"NewClass\" ]"
```
Otto-AA, line 622: I'm surprised that most type checkers understand nested quotes. Pyrefly currently fails on this (https://github.com/facebook/pyrefly/issues/3006). If I find the time, I'll check if we can get rid of the stringification by moving the mutated methods after the class body and inside the class body add a trampoline like we do for other class methods (the method that looks like the normal method but passes the args to `_mutmut_trampoline(...)`.

tests/threading/test_timeout.py
```diff
@@ -0,0 +1,173 @@
...
+import pytest
+
+from mutmut.threading import timeout
+
+
+@pytest.fixture(autouse=True)
+def clear_heap():
+    timeout._timeout_heap.clear()
+    yield
+    timeout._timeout_heap.clear()
```
Otto-AA, line 16: I like this safety fixture ❤️

tests/mutation/test_pragma_handling.py
```diff
@@ -0,0 +1,352 @@
...
+z = 3
+"""
+        no_mutate, ignore_node_lines = parse_pragmas("test.py", source)
+        assert no_mutate == set()
+        assert ignore_node_lines == {2, 3, 4}
+
+    def test_does_not_affect_code_after_dedent(self):
+        source = """
+def foo():
+    # pragma: no mutate block
```
Otto-AA, line 150: It's not clear to me from the tests, if `# pragma: no mutate block` also prevents mutations previously to this comment. e.g.:

```python
def foo():
  x = 1
  # pragma: no mutate block
  y = 2
```

I'd add a test case for this, so we don't modify the behaviour in the future. Either behaviour would be fine for me I think.

src/mutmut/mutation/pragma_handling.py
```diff
@@ -0,0 +1,192 @@
...
+        return None
+    tail = text.partition("no mutate")[-1].strip()
+    tokens = tail.lstrip(": ").split(",", 1)[0].split()
+    tok = tokens[0] if tokens else None
+    if tok in ("block", "start", "end"):
+        return tok
+    return "bare"
+
+
+class PragmaVisitor(cst.CSTVisitor):
```
Otto-AA, line 30: I think it could be good for performance, if we add some heuristic to skip visiting all nodes if eg there is no `# pragma` or `no mutate` inside the source code string. I'll test it.

src/mutmut/mutation/file_mutation.py
Otto-AA: I'll remove stringification in a follow-up commit on the main branch by putting the mutated methods below the class body.

tests/mutation/test_pragma_handling.py
Otto-AA: I'll add a test in a follow-up commit.

src/mutmut/mutation/pragma_handling.py
Otto-AA: I did not see a significant slowdown compared to the main branch, so let's not overcomplicate.

src/mutmut/mutation/file_mutation.py
Otto-AA: I've tried to fix this, but ended up disabling enums and staticmethod/classmethod mutation for now if type_checking is used. I've added an E2E test with some cases that would need to be fixed, maybe there's even more. Definitely not trivial.

src/mutmut/mutation/file_mutation.py
Otto-AA: Thanks for highlighting this issue. I'll remove the stringification in a follow-up commit, as it seems a bit brittle overall and we can work around it without much trouble.

nicklafleur:
@Otto-AA I appreciate the review and additions. I still have some changes to put up once I'm rebased on top of main but feel no pressure. Next pr is function hashing and dependency tracking so by no means do I expect a speedy turn around, take all the time you need to balance things out in a more sustainable way :). 

Thanks for all you do for OSS <3

PR #478, opened 2026-03-06
Add mutation for ternary operators

### LINKED ISSUE #196: Add mutation for ternary operators
Ternary operator errors of not covering both branches can't be detected by coverage, so it would be useful to mutate these to enable/disable each branch.  E.g.

```python
mark = "x" if count % 2 else "o"
```

mutation into:
```python
mark = "x" if count % 2 and False else "o"
````
and

```python
mark = "x" if count % 2 or True else "o"
````

This types of mutations aren't as useful for normal if statements, because branch coverage can detect when one of the alternatives isn't covered.

### PR DESCRIPTION
**Pull request by ryanfreckleton**
Closes #196

This adds mutations for ternary expressions (`a if cond else b`), as discussed in the issue.

The new mutations target the `IfExp` condition and force each branch explicitly:

```python
mark = "x" if count % 2 else "o"
```

can now mutate to:

```python
mark = "x" if count % 2 and False else "o"
mark = "x" if count % 2 or True else "o"
```

Implementation notes:
- adds an `IfExp` mutation operator in `node_mutation.py`
- uses LibCST's `IfExp` node, following the implementation pointer in the issue comments
- wraps low-precedence conditions like `b or c` before appending `and False` / `or True`, so we preserve the original condition semantics

I kept this scoped to ternary operators only, not normal `if` statements, which matches the issue rationale that branch coverage is less helpful for inline conditionals.

Tests:
- added direct mutation tests for ternary expressions
- added a precedence-preservation test for boolean conditions inside ternaries
- updated the existing mutated-function assertion to check for the new ternary mutants without depending on exact mutant numbering

### DIFF
--- a/src/mutmut/node_mutation.py
+++ b/src/mutmut/node_mutation.py
@@ -242,6 +242,23 @@ def operator_assignment(
     yield node.with_changes(value=mutated_value)
 
 
+def _wrap_low_precedence_expression(expression: cst.BaseExpression) -> cst.BaseExpression:
+    """Wrap low-precedence expressions before appending boolean operators."""
+    if expression.lpar or expression.rpar:
+        return expression
+
+    if isinstance(expression, cst.BooleanOperation | cst.IfExp | cst.Lambda | cst.NamedExpr):
+        return expression.with_changes(lpar=[cst.LeftParen()], rpar=[cst.RightParen()])
+
+    return expression
+
+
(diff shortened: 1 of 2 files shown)


nicklafleur:
@ryanfreckleton since there's been no activity on this PR in 2 months, I'm going to go ahead and close it. The contribution and idea seems really interesting, and I would encourage you to re-submit an updated PR for it.

PR #479, opened 2026-03-10
🐛 Change cwd for coverage reports

sam-mosleh:
First time installing `mutmut` — excited to try it out, but ran into a couple of issues on my machine that blocked me from running it at all on my MacBook running Tahoe v26.3.1. This PR fixes both:

## ~~Fix 1: Segfault on startup due to `setproctitle`~~

~~On macOS, `mutmut` crashes immediately with a segfault. After some digging, the culprit is the call to `setproctitle` which fails simply because **forking on macOS is inherently unsafe**. I want to be upfront that **this may not be the right long-term fix**. I went with the simplest change to unblock myself, but I'm very open to guidance from you (@boxed) on the preferred approach.~~

## Fix 2: Coverage configuration not being respected

Even after unblocking the segfault, I noticed that coverage reports were not picking up the project's coverage configuration (e.g., `.coveragerc` or `[tool.coverage]` in `pyproject.toml`). For example omit configs fail to locate files because it is called from the parent `mutants` directory.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,36 +1,36 @@
 default_language_version:
-  python: python3.14
+    python: python3.10
 repos:
-  - repo: https://github.com/pre-commit/pre-commit-hooks
+-   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v6.0.0
     hooks:
-      - id: check-docstring-first
+    -   id: check-docstring-first
         exclude: ^e2e_projects/
-      - id: check-json
-      - id: check-merge-conflict
+    -   id: check-json
+    -   id: check-merge-conflict
         exclude: \.rst$
-      - id: check-yaml
-      - id: debug-statements
+    -   id: check-yaml
+    -   id: debug-statements
         exclude: tests/data/test_generation/invalid_syntax\.py
-      - id: end-of-file-fixer
-      - id: trailing-whitespace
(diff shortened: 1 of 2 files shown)


src/mutmut/__main__.py
```diff
@@ -643,6 +649,19 @@ def run_tests(self, *, mutant_name: str | None, tests: Iterable[str]) -> int:
...
+            self.prepare_main_test_run()
+            pytest_args = [
+                "-x",
+                "-q",
+                "-p",
+                "no:randomly",
+                "-p",
+                "no:random-order",
+            ] + self._pytest_add_cli_args_test_selection
+            return int(self.execute_pytest(pytest_args))
```
Otto-AA: I think should extract a common function between `run_tests` and `collect_main_test_coverage`, such that we don't get unexpected differences between their implementations in the future.

.pre-commit-config.yaml
```diff
@@ -1,36 +1,36 @@
 default_language_version:
-    python: python3.10
+  python: python3.14
```
Otto-AA: The default local python version is 3.10 (also specified in .python-version). I think we should not change this.

sam-mosleh (commented):
Thanks for the review @Otto-AA

src/mutmut/__main__.py
sam-mosleh: I'll have to refactor other places too then

PR #481, opened 2026-03-21
fix ruff deprecation in pre-commits

even-even:
Hello
In this PR:
1) fix ruff legacy alias https://github.com/astral-sh/ruff-pre-commit/blob/main/.pre-commit-hooks.yaml#L23
2) sync ruff versions in dependencies and pre-commits
3) add https://docs.astral.sh/ruff/settings/#show-fixes flag


Otto-AA:
Hi, thanks for the PR, looks good to me.

In case there's another dependency to update in the future, I'd prefer issues over PRs (verifying uv.lock changes is a bit tedious; doing it manually is faster)

PR #482, opened 2026-03-23
fix: guard against empty mutants table in ResultBrowser (browse command)

m42e:
## Problem

Running `mutmut browse` crashes with `RowDoesNotExist: Row index 0 is not valid.` when pressing `t` (view tests) or triggering other mutant actions before a file with mutants is selected.

The root cause: `DataTable.cursor_row` returns `0` even when the table is empty, so the existing `cursor_row is None` guard does not protect against calling `get_row_at(0)` on an empty table.

## Fix

Use `is_valid_row_index()` (provided by Textual's `DataTable`) in addition to the `None` check in both `get_mutant_name_from_selection` and `action_apply_mutant`.


boxed:
thanks

PR #483, opened 2026-03-23
fix: guard set_start_method against duplicate calls (GH-466)

### LINKED ISSUE #466: [BUG] Context already been set
### Description of the issue

When running mutmut3 I always end up with the same error during the stats step: 

```
RuntimeError: context has already been set
```

During the stats step on the first 

```python
from mutmut.__main__ import record_trampoline_hit
```

`mutmut.__main__` is loaded and so `set_start_method('fork')` is executed a second time.

Whatever the project I use, as long as I use mutmut >3, I always get that error. I've tried even with a dummy project (CRUD of FastAPI), and always end up with the same issue. I'm out of ideas of what to do, so I'm asking here if anyone has an idea of what is happening here.

### Specs

OS: MacOS 15.7 (M4 pro)
Python: 3.13, 3.14
mutmut: 3.4.0
pytest: 9.0.2
Project layout: flat

tomrussobuilds:
Fixes #466

When mutmut is invoked via the entry-point wrapper (`mutmut run`),
`mutmut.__main__` is cached in `sys.modules` and not re-executed.
When invoked via `python -m mutmut run`, the module is re-executed,
causing `set_start_method("fork")` to raise `RuntimeError: context
has already been set`.

Guard with `get_start_method(allow_none=True)` and emit a warning
when the module is re-executed.


Otto-AA:
Thank you for the PR, I've resolved the merge conflicts and add the commit here: https://github.com/boxed/mutmut/commit/22e26fb549e2f0e3cdfcfbebcbe2f3aa572b27a4

PR #484, opened 2026-03-25
🐛 Fix mypy strict mode compatibility

sam-mosleh:
## Problem

When running mypy in strict mode, the following exception is raised:

```
Exception: Could not find mutant for type error .../__init__.py:10 (Unused "type: ignore" comment). Probably, a code mutation influenced types in unexpected locations. If your project normally has no type errors and uses mypy/pyrefly, please file an issue with steps to reproduce on github.
```

## Changes

This PR adds a new e2e test that runs mypy in strict mode to catch this class of error early and prevent future regressions. Additionally, the Docker image has been restructured to separate dependency installation from test files — previously, dependencies were being rebuilt on every single run because test files were updated during each run, invalidating the layer cache; now dependencies are built once and subsequent runs simply execute the tests against the pre-built image.


sam-mosleh:
@Otto-AA Can I remove the space from `tests/test_mutation regression.py` filename ?

Otto-AA:
Would disabling the error work for you? I'm not using mypy a lot, but I guess `--disable-error-code unused-ignore` should work.

I won't have time to look at this in the near future. Also, I think using `# type: ignore` excessively for simpler mutmut code could be a reasonable tradeoff.

sam-mosleh:
@Otto-AA Totally understand the bandwidth constraints so no pressure on timeline.
I do think `--disable-error-code unused-ignore` puts the burden on the user in a way that's easy to miss, especially since the error surface isn't obvious. Someone running mutmut against a strict-mode mypy project is likely to hit the exception cold with no clear path forward.
If merging the full PR isn't feasible right now, even a brief mention in the docs that strict-mode mypy users may need to disable `unused-ignore` would go a long way.
I'm happy to put together a minimal docs-only PR if that's a lower-lift alternative.

Otto-AA:
I've added it to the README, makes sense to inform users early on: https://github.com/boxed/mutmut/commit/2644bb5a862c85cad3ab9091e3fa17ed5cb03431

And I've also added E2E tests for mypy strict in #500 (and added a `# type: ignore` for the trampoline setup which seemed necessary for `mypy --strict` with methods that return `None`).

PR #487, opened 2026-03-29
fix: support non-editable installs (site-packages import precedence)

pythoniste:
When the package-under-test is installed in site-packages as a
  non-editable wheel rather than via `pip install -e .`, mutmut's
  trampoline-injected code in the mutants/ directory was never loaded.

  Python imported the original (non-mutated) modules from site-packages
  because they were already cached in sys.modules. The stats phase found
  zero test coverage and all mutants were marked "not checked".

  Add _patch_imported_packages() to setup_source_paths(): after adding
  mutants/ to sys.path, patch __path__ on every already-imported package
  so submodule lookups prefer the mutants tree. Flush leaf modules whose
  mutated .py exists on disk so they get re-imported with trampoline code.

  Add _copy_parent_init_files() to copy_src_dir(): when paths_to_mutate
  targets individual files, parent __init__.py files were missing from
  the mutants tree, preventing Python from recognising it as a package.

  Tested on a Pydantic-heavy project (8,996 mutants): stats phase now
  finds 477 covered functions and achieves a 70% kill rate, where
  previously it found zero coverage.

### DIFF
--- a/src/mutmut/__main__.py
+++ b/src/mutmut/__main__.py
@@ -390,6 +390,31 @@ def _has_mutant_file(mod_name: str, mutated_packages: dict[str, str]) -> bool:
     return candidate.exists()
 
 
+def _flush_test_modules() -> None:
+    """Remove cached test modules from sys.modules so they get re-imported.
+
+    Called in the forked child process when testing a mutant that targets
+    import-time code (e.g. __init_subclass__).  Without this, the child
+    inherits cached modules from the parent (stats phase) and
+    __init_subclass__ never fires again with the mutant active.
+
+    Only non-package modules under ``tests`` (or ``conftest``) are flushed.
+    The mutated package itself stays cached — its trampolines dynamically
+    check ``MUTANT_UNDER_TEST`` on each call.
+    """
+    to_remove = [
+        name
+        for name in sys.modules
+        if name == "conftest"
+        or name.startswith("conftest.")
+        or name.startswith("tests.")
+        or name.startswith("tests_")
+    ]
+    for name in to_remove:
+        del sys.modules[name]
+    importlib.invalidate_caches()
+
+
 def store_lines_covered_by_tests() -> None:
     assert mutmut.config is not None
     if mutmut.config.mutate_only_covered_lines:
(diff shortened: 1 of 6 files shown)


pythoniste:
By the way, the branch name is misleading. At first, I wanted to fix Pydantic and then, I fixed something else.

src/mutmut/__main__.py
```diff
@@ -213,6 +214,41 @@ def copy_src_dir() -> None:
...
+def _copy_parent_init_files() -> None:
+    """Copy ``__init__.py`` for every package directory in the mutants tree.
+
+    When ``paths_to_mutate`` targets individual files (e.g.
+    ``["pkg/sub/module.py"]``), mutmut creates ``mutants/pkg/sub/module.py``
+    but does *not* copy the ``__init__.py`` files in ``pkg/`` or ``pkg/sub/``.
+    Without them Python's import system cannot recognise ``pkg`` as a package
+    and falls back to the copy installed in site-packages, skipping the
+    trampoline-injected mutant code entirely.
+    """
```
Otto-AA, line 232: paths_to_mutate should not cover individual files (#414 )

src/mutmut/__main__.py
```diff
@@ -266,17 +302,118 @@ def create_file_mutants(path: Path) -> FileMutationResult:
...
+    else:
+        candidate = Path(root, parts[0] + ".py")
+    return candidate.exists()
+
+
+def _flush_test_modules() -> None:
+    """Remove cached test modules from sys.modules so they get re-imported.
+
+    Called in the forked child process when testing a mutant that targets
+    import-time code (e.g. __init_subclass__).  Without this, the child
```
Otto-AA, line 397: I'm a bit confused, mutmut does not support import-time code mutations. Is this a feature PR to add this?

e.g. in following file:

```python
MY_CONST = 1234

def foo():
  return 1234
```

Only `foo` will be mutated, `MY_CONST` won't be mutated.

Otto-AA (commented):
Hi, thanks for the PR.

I don't have a lot of time for reviewing currently, so I probably won't get to test and read through this PR in the next weeks.

I don't fully understand the editable vs non-editable problem yet. Why does the editable install work fine, but the non-editable version gets imported as non-mutated version and needs to get flushed out of the modules? We do `sys.path.insert(0, abs_path)` to preferably import mutated code, is the non-editable version already installed at this point?

And would installing the project as editable be possible for you? I would tend to detect if the project is installed as non-editable and raise an Exception asking the user to install as editable instead.

nicklafleur:
@pythoniste Just checking if this work is still active and if you intend on rebasing on top of main. If there's no response in the next ~week I'll go ahead and close the PR

PR #491, opened 2026-04-04
Follow up commits

Otto-AA:
Follow up of #476 

Most importantly, do not mutate staticmethod, classmethod and enums when type_checking is enabled.

@nicklafleur In case you disagree with any of these changes, let me know and we can adapt in a future commit.


nicklafleur:
@Otto-AA thanks for the updates, had some company over the weekend but will check these out and explore the type_checking problem a bit better to gain a better understanding of what the conflict between the two is, any information on the topic is more than welcome of course.

PR #494, opened 2026-04-10
fix: external trampoline type checking

nicklafleur:
Addresses issues introduced in https://github.com/boxed/mutmut/pull/476

Changes:
- Detects the `Self` annotation and replaces it with generics in mutated functions to support type checking
- Normalize external trampoline naming conventions to the established patterns (this had some side-effects with type checking as well)

Tests:
- add tests that validate inheritence of class/static methods passes type checking with the generic replacement

Ref: https://github.com/boxed/mutmut/commit/c8666814cecbdc5bd39d96e3b8dee07c89dc499d

```python
class Color(Enum)
    @classmethod
    def create(cls, name: str) -> Self:
        return cls(name)
...
_TxǁColorǁ = TypeVar('_TxǁColorǁ', bound='Color')

def xǁColorǁcreate__mutmut_orig(cls: type[_TxǁColorǁ], name: str) -> _TxǁColorǁ :
    return cls(name)
```


Otto-AA:
Thanks for the PR, works on my sample project and looks typing conform (https://typing.python.org/en/latest/spec/generics.html#use-in-classmethod-signatures).

I didn't know that `Self` is more or less a wrapper for a TypeVar, good approach :)

I've rebased this commit and merged it in #502

PR #499, opened 2026-04-15
fix: avoid Python name mangling in trampoline prefix for underscore-prefixed classes

### LINKED ISSUE #498: NameError: trampoline name mangling breaks classes with leading underscore
# NameError: trampoline name mangling breaks classes with leading underscore

## Description

Classes with a leading underscore (e.g. `_MyValidator`) cause a `NameError` at import time because the generated trampoline name triggers Python's [name mangling](https://docs.python.org/3/reference/expressions.html#atom-identifiers).

```
NameError: name '_CrossDomainValidator__CrossDomainValidator_validate_trampoline' is not defined
```

This prevents the entire test suite from running - not just the affected mutant.

## Root Cause

`trampoline_templates.py` (L70) and `file_mutation.py` (L364) both build the prefix as:

```python
prefix = f"_{class_name}_{method_name}"
```

When `class_name = "_CrossDomainValidator"`, the resulting name starts with `__`, which Python mangles inside the class body. The module-level function keeps the original name, so the reference breaks.

## Suggested Fix

Strip leading underscores from `class_name` in the prefix:

```python
prefix = f"_{class_name.lstrip('_')}_{method_name}"
```

**Caveat:** if a module contains both `class Foo` and `class _Foo` with the same method name, this would produce identical prefixes. An alternative that avoids any collision risk:

```python
prefix = f"_mutmut_{class_name}_{method_name}"
```

## Real-World Case

I hit this in [orchard-ml](https://github.com/tomrussobuilds/orchard-ml) on the `_CrossDomainValidator` class in [`orchard/core/config/manifest.py`](https://github.com/tomrussobuilds/orchard-ml/blob/main/orchard/core/config/manifest.py). The `lstrip` fix resolves the issue - mutmut runs and all mutants are processed correctly.

Happy to open a PR with either approach.

tomrussobuilds:
Use '_mutmut_' prefix instead of '_{class_name}_' to prevent
identifiers starting with '__' inside class bodies, which Python
mangles and causes NameError at import time.

Fixes #498

### DIFF
--- a/src/mutmut/mutation/file_mutation.py
+++ b/src/mutmut/mutation/file_mutation.py
@@ -361,7 +361,7 @@ def _external_method_injection(
     external_nodes: list[MODULE_STATEMENT] = []
     mutant_names: list[str] = []
     method_name = method.name.value
-    prefix = f"_{class_name}_{method_name}"
+    prefix = f"_mutmut_{class_name}_{method_name}"
     mangled_name = mangle_function_name(name=method_name, class_name=class_name) + "__mutmut"
 
     orig_func = method.with_changes(name=cst.Name(f"{prefix}_orig"), decorators=[])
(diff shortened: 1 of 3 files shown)


boxed:
Thanks

PR #502, opened 2026-04-18
fix: external trampoline type checking (rebased)

Otto-AA:
#494 rebased


nicklafleur:
oh sorry about not rebasing! Had a hackathon at work these past two weeks so my mind has been elsewhere, had this on my list for next week!

Thanks for the rebase :)

PR #513, opened 2026-05-03
fix: only count user-code frames toward max_stack_depth

### LINKED ISSUE #378: Only count user-code frames for `max_stack_depth`
Currently, we also count frames created by mutmut to apply `max_stack_depth` (in particular `_mutmut_trampoline`).  As such, users cannot really know what the number means, or which number to choose.

I'd suggest to filter out frames:
- that are created by mutmut
- that are outside of the `./mutants` folder (e.g. 3rd party packages)

irajaryan:
Closes #378.

`record_trampoline_hit` previously decremented the `max_stack_depth` budget for every frame walked, including its own frame, the injected `_mutmut_trampoline`, and any third-party / stdlib frames. This meant a configured value of N didn't correspond to anything a user could reason about.

This change only counts frames whose file lives under the `mutants/` working directory and whose function name is not `_mutmut_trampoline`. `os.path.realpath` is used so the comparison works on macOS where `/var` is a symlink to `/private/var`.

The e2e `config` project's `max_stack_depth` is updated from 8 to 5 to keep the same coverage pattern under the new counting; the snapshot is unchanged.

Six unit tests cover: budget disabled, within/over budget, third-party frames excluded, trampoline frames excluded, and the existing test-runner heuristic preserved.

### DIFF
--- a/e2e_projects/config/pyproject.toml
+++ b/e2e_projects/config/pyproject.toml
@@ -28,7 +28,7 @@ source_paths = [ "config_pkg/" ]
 only_mutate = [ "config_pkg/logic/*" ]
 do_not_mutate = [ "*ignore*" ]
 also_copy = [ "data" ]
-max_stack_depth=8 # Includes frames by mutmut, see https://github.com/boxed/mutmut/issues/378
+max_stack_depth=5 # Counts only user-code frames; see https://github.com/boxed/mutmut/issues/378
 # verify that we can override options with pytest_add_cli_args
 pytest_add_cli_args = ["-o", "xfail_strict=False"]
 # verify test exclusion (-m 'not fail') and test inclusion (-k=test_include)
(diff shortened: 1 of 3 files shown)


Otto-AA:
Thanks for the PR, I didn't see it and worked on this simultaneously together with some other changes in #514 . There I decided to only include stack frames part of the `source_paths` config, so test setup is also not counted.

> os.path.realpath is used so the comparison works on macOS where /var is a symlink to /private/var.

That's something I did not think about in my implementation. Resolving symlinks could indeed be necessary.

irajaryan:
> Thanks for the PR, I didn't see it and worked on this simultaneously together with some other changes in #514 . 

Ah, didn't see #514, looks like a much bigger cleanup and I think filtering against source_paths is the right call (it correctly excludes test files, which my approach doesn't).

Happy to close this in favor of #514. Let me know if I need to open a follow-up PR with just the realpath fix once #514 lands, or if you'd prefer to fold it into #514 directly. The relevant bit is small.

Otto-AA:
I've changed my implementation to use [Path.resolve](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve), which also resolves symlinks. I think it should cover the same use cases as os.path.realpath. If there's something not working, please let me know or create a PR.

And thank you for the PR nonetheless! I'm still surprised that we both decided at the same time to fix this, after the issue has been open for a year already

Issue #416, opened 2025-07-19
Mutants not tested with pythonpath or sys.path.append


CyberiaResurrection:
Mutmut version: 22f3924b6b5a2ff7916c1f62aa1ed1f3341e42ee
Python version: 3.9.21
OS: Fedora Linux 41

Repo: https://github.com/makhidkarun/traveller_pyroute

pyproject.toml bits:
```
paths_to_mutate = ["PyRoute/"]
tests_dir = ["Tests/"]
also_copy = ["Tests/", "PyRoute/Pathfinding/*.so"]
runner = "pytest -x"
```

After getting past #415 , I'm now actually _getting_ mutants to run, but:
```
⠼ Generating mutants
    done in 44193ms
⠇ Listing all tests 
⠼ Running clean tests
    done
⠧ Running forced fail test
    done
Running mutation testing
⠴ 30097/30097  🎉 0 🫥 30097  ⏰ 0  🤔 0  🙁 0  🔇 0
0.00 mutations/second

```

no mutants are actually being tested.
I'm not sure how this case differs from #351 , but the following are things I can find:
1.  **mutants/** directory not committed to git.
2. **Tests/** directory is not being mutated.
3. In the branch I'm working in, I've removed _PyRoute._ from imports, across both code and tests.
4. Forced-fail testing runs clean.

Otto-AA:
What did you change with regards to the previous issue? In the previous issue the mutants were tested and it was looking okay (except for the additional segmentation fault logging, which should be shown as emoji).

CyberiaResurrection:
Ah, I see.  After fixing the segfault-leading code, I ran mutation again and got (see above)
```
⠴ 30097/30097  🎉 0 🫥 30097  ⏰ 0  🤔 0  🙁 0  🔇 0
```
We were looking at different status lines.  Now to try debug = true, and also running pytest from mutants folder

Otto-AA:
> FAILED: Unable to force test failures

This usually indicates a problem with the project/test setup and imports. Mutmut copies all relevant source and test files to `./mutants` and mutates the source files there. The "force test failures" check sets `os.environ['MUTANT_UNDER_TEST'] = 'fail'` which will cause all mutants to raise a `MutmutProgrammaticFailException`. This tests if the mutants can make any test fail (and thus the setup looks fine).

In your case, something in the setup did not work and all tests passed, despite mutmut trying to make all mutants fail.

[...]

Issue #351, opened 2024-12-06
Mutmut doesn't test mutants


GianvincenzoConcia:
I have a problem with mutmut: the tool creates mutants but it doesn't tests them.

Mutants using command mutmut browse:
![image](https://github.com/user-attachments/assets/ad917ba2-4425-4a2d-9f0b-75746cbd52db)

Results:
![image](https://github.com/user-attachments/assets/a74a541d-3f14-4371-b509-56817f492d01)

what could be the problem??

@boxed

boxed:
Did you run `mutmut run` first? If so, what was the output?

GianvincenzoConcia:
![image](https://github.com/user-attachments/assets/811680bb-d1e0-4488-b6ec-2aa0db81b2da)

GianvincenzoConcia:
This is the output:
```
mutmut run
⠹ Generating mutants
    done in 4ms
⠧ Listing all tests 
⠸ Running clean tests
    done
⠧ Running forced fail test
    done
Running mutation testing
⠇ 380/380  🎉 0 🫥 380  ⏰ 0  🤔 0  🙁 0  🔇 0
0.00 mutations/second
```

boxed:
Is this project available somewhere so I can try this?

vinceslao:
https://github.com/vinceslao/CitationAnalyzer

vinceslao:
excuse me but what is the reason?

boxed:
They are temporary files for working with mutation testing. I bet it's the same with the cosmic ray files you have committed. At least the .sqlite files seem iffy to commit. 

You have also committed `.DS_Store` incorrectly I se.

vinceslao:
ok, I fixed that. Any updates on the initial issue?

MrMicc:
I'm facing something similar issue. Mutants are being created, but the tests are not being executed 

Python version: `3.12.5`

setup.cfg
```yaml
[mutmut]
paths_to_mutate = src/
runner = pytest tests/
tests_dir = tests/
```
Also tested without `runner` at setup.cfg file

result `mutmut run`:
```
⠦ Generating mutants
    done in 5ms
⠋ Listing all tests 
⠴ Running clean tests
    done
⠦ Running forced fail test
    done
Running mutation testing
⠇ 51/51  🎉 0 🫥 51  ⏰ 0  🤔 0  🙁 0  🔇 0
0.00 mutations/second
```

result of `mutmut results`:
```
[...]

Issue #341, opened 2024-11-03
All the statuses become “not checked”


32145072:
Hello.
I am currently learning about mutation testing .

I tried it after mutmut 3.x was released, but I am having trouble solving the problem where all the statuses become “not checked”.

When I ran `mutmut run`, I got an error saying `failed to collect stats, no active tests found`.

The results of the` mutmut browse` are as follows.
![mutmut1](https://github.com/user-attachments/assets/330e8fa8-5b7c-4c2b-bfe8-af17587c34d7)

The generated meta file is as follows.
```
{
    "exit_code_by_key": {
        "fizz_buzz.x_fizz_buzz__mutmut_1": null,
        "fizz_buzz.x_fizz_buzz__mutmut_2": null,
        "fizz_buzz.x_fizz_buzz__mutmut_3": null,
        "fizz_buzz.x_fizz_buzz__mutmut_4": null,
        "fizz_buzz.x_fizz_buzz__mutmut_5": null,
        "fizz_buzz.x_fizz_buzz__mutmut_6": null,
        "fizz_buzz.x_fizz_buzz__mutmut_7": null,
        "fizz_buzz.x_fizz_buzz__mutmut_8": null,
        "fizz_buzz.x_fizz_buzz__mutmut_9": null,
        "fizz_buzz.x_fizz_buzz__mutmut_10": null,
        "fizz_buzz.x_fizz_buzz__mutmut_11": null,
        "fizz_buzz.x_fizz_buzz__mutmut_12": null,
        "fizz_buzz.x_fizz_buzz__mutmut_13": null,
        "fizz_buzz.x_fizz_buzz__mutmut_14": null,
        "fizz_buzz.x_fizz_buzz__mutmut_15": null
    },
    "hash_by_function_name": {
        "fizz_buzz": "b62123013949868a4a537fdf38533c9c"
    }
}
```

I also tried running it on other projects such as astmonkey, but I was unable to see the correct  status.
Is there any way to solve this?

About my system: (Please let me know if there is any information missing.)
1. Python version - Python 3.9.20
2. OS - Windows 10
3. wsl - Ubuntu 24.04.1
4. mutmut ver - 3.2.0

```setup.cfg
[mutmut]
paths_to_mutate=src/
tests_dir=tests/
```
```pytest.ini
[pytest]
pythonpath = src
```

Thank you.

boxed:
It looks like this project is something you can share with me to take a look at?

32145072:
[...]

PR #516, opened 2026-05-08
fix: diagnose silent test-to-mutant key mismatch in stats

### LINKED ISSUE #515: All mutants reported as "No Tests" when test imports differ from the file-path module name
Hit this on a real codebase and reduced to a minimal repro. mutmut runs to completion, every mutant is marked "No Tests", but the test suite passes when run directly with pytest and the trampolines actually execute. The early-bail check in `run_stats_collection` doesn't catch it because hits *are* being recorded during stats collection, just under keys that no mutant lookup will ever match.

### Repro

```
buggy/
├── pyproject.toml
├── pkg/
│   ├── __init__.py
│   └── foo.py        # def add(a, b): return a + b
└── tests/
    └── test_foo.py   # import foo;  def test_add(): assert foo.add(2, 3) == 5
```

```toml
[tool.mutmut]
source_paths = ["pkg"]

[tool.pytest.ini_options]
pythonpath = ["pkg"]
```

- `pytest -q` -> 2 passed
- `mutmut run` -> `3/3  🎉 0 🫥 3 ⏰ 0 🤔 0 🙁 0 🔇 0 🧙 0`, 0.00 mutations/second

### What's happening

After patching prints into `pytest_runtest_teardown` and `mangled_name_from_mutant_name`, the recorded and lookup keys come out as:

```
RECORD test='tests/test_foo.py::test_add'         hits=['foo.x_add']
RECORD test='tests/test_foo.py::test_is_positive' hits=['foo.x_is_positive']
LOOKUP mutant='pkg.foo.x_add__mutmut_1'           key='pkg.foo.x_add'
LOOKUP mutant='pkg.foo.x_is_positive__mutmut_1'   key='pkg.foo.x_is_positive'
```

When the test does `import foo`, `__module__` is `'foo'`, so `record_trampoline_hit` writes `'foo.x_add'`. The mutant name is built from the file path `pkg/foo.py`, so the lookup key is `'pkg.foo.x_add'`. They never match.

The early-bail check around `__main__.py:736` sums total recorded associations (2 here), which passes the `> 0` gate, so the run continues and every per-mutant lookup quietly returns the empty set.

### Workaround

Use qualified imports in tests (`from pkg.foo import add`) and drop the `pythonpath` injection. For repos with hyphenated source dirs, also rename to underscore and switch internal imports to relative.

irajaryan:
This adds a sanity check after stats collection that detects when tests recorded trampoline hits but none match any mutant lookup key. Without it, the run silently completes with every mutant marked "No Tests" and 0.00 mutations/second.

The trampoline records `orig.__module__ + '.' + orig.__name__` while per-mutant lookup uses the path-derived dotted name from `get_mutant_name`. They match in healthy projects (qualified package imports, source under `./`, `src/`, or `source/`). They diverge when tests use top-level imports enabled by pytest's `pythonpath` config, conftest `sys.path.insert`, or non-conventional source dirs. The existing zero-check in `run_stats_collection` doesn't catch the divergence because hits are recorded, just under unmatchable keys.

The new `_check_test_to_mutant_associations` runs after `collect_source_file_mutation_data` and exits early with sample recorded vs expected keys, in the same style as the existing "Stopping early" diagnostic.

Same symptom appears in #341, #351, #416, #419, all closed after users fixed their project setup. This change surfaces the cause earlier.

Fixes #515

### DIFF
--- a/src/mutmut/__main__.py
+++ b/src/mutmut/__main__.py
@@ -881,6 +881,54 @@ def collect_source_file_mutation_data(
     return mutants, source_file_mutation_data_by_path
 
 
+def _check_test_to_mutant_associations(
+    source_file_mutation_data_by_path: dict[str, SourceFileMutationData],
+) -> None:
+    """Detect when stats recorded trampoline hits but no recorded key matches any mutant key.
+
+    The trampoline records ``orig.__module__ + '.' + orig.__name__`` while the
+    per-mutant lookup uses the path-derived dotted name from ``get_mutant_name``.
+    In a healthy project these are equal because the test suite imports the
+    source via fully-qualified package paths and the source dir matches one of
+    the conventions handled by ``setup_source_paths``. When they diverge (e.g.
+    ``[tool.pytest.ini_options] pythonpath = ["pkg"]`` makes tests do
+    ``import foo`` instead of ``from pkg.foo import ...``) every mutant is
+    silently marked "No Tests" and the run reports 0.00 mutations/second.
+
+    This check exits with an actionable message instead of producing the silent
+    all-No-Tests outcome.
+    """
+    recorded = set(mutmut.tests_by_mangled_function_name.keys())
+    if not recorded:
+        # No hits at all - the existing zero-check in run_stats_collection
+        # already covers this path; nothing to add here.
+        return
+
+    expected = {
+        mangled_name_from_mutant_name(mutant_name)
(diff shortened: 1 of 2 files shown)


Otto-AA:
Thanks for the PR, I like the proactive warning. I guess we should be able to fix this within mutmut, but until then the warning will help :)

Issue #486, opened 2026-03-29
mutmut v3 copies source to temp dir, breaking editable-install imports


dirkenglund:
## Description

When using mutmut v3 (3.5.0) on a project installed with `pip install -e .`, mutations fail because mutmut copies source files to a temporary directory where the editable package's import paths no longer resolve.

## Reproduction

**Project structure:**
```
my-project/
├── pyproject.toml          # [project] name = "my_project"
├── src/
│   └── my_project/
│       ├── __init__.py
│       └── core.py         # file to mutate
└── tests/
    └── test_core.py        # imports my_project.core
```

**Steps:**
1. `pip install -e .`
2. `pytest tests/` — passes
3. `mutmut run --paths-to-mutate src/my_project/core.py`
4. Mutations fail with `ModuleNotFoundError: No module named 'my_project'`

## Expected behavior

Mutations should run in-place (or with sys.path adjusted) so that editable installs and relative imports continue to work.

## Actual behavior

mutmut copies files to `/tmp/` (or a `mutants/` directory), then runs pytest from that context. The editable install's `.pth` file points to the original `src/` directory, so the mutated code is never imported — or worse, the original un-mutated code is imported instead, producing false "survived" results.

## Workaround

Manual mutation testing via subprocess + file patching (mutate in-place, run tests, restore original).

## Related

- #456 reports a similar import path issue with Django projects
- #485 may also be related (initial test run failure)

## Environment

- mutmut 3.5.0 (pip reports 2.4.5, `mutmut.__version__` reports 3.5.0)
- Python 3.11.9
- macOS 14 (arm64)
- Project uses `src/` layout with pyproject.toml

boxed:
The mutmut run command is a bit strange. That looks like a mutmut 2.x style invocation, not what mutmut 3.x does. What happens if you run just `mutmut run` like the docs say?

Otto-AA:
Can you create a github repo with the minimal example? I could not reproduce it locally.

> pip reports 2.4.5

This sounds suspicious 🤔

raschmitt:
## Related failure: `copy_src_dir` crashes on socket files in CI

We hit a different manifestation of the same `copy_src_dir` issue on GitHub Actions (Ubuntu runner, Python 3.12).

**Error:**
```
FileNotFoundError: [Errno 2] No such file or directory: '/run/user/1001/systemd/units/invocation:gpg-agent-ssh.socket'
```

**Full traceback:**
```
[...]

PR #517, opened 2026-05-09
fix: ignore sockets and other non-file files

Otto-AA:
Addresses https://github.com/boxed/mutmut/issues/486#issuecomment-4408940704


boxed:
This looks like a straight forward change, why is it marked as draft?

Otto-AA:
I don't really remember, but I think (1) because I'm not sure if there are any non-regular files we *should* include and (2) because I didn't verify if mutmut works correctly with this change, and it resolves the socket crash (e.g. because I missed a change in some other part)

Issue #526, opened 2026-06-10
`record_trampoline_hit` resolves `source_paths` against the cwd on every hit, aborting collection for any `chdir`-ing test


joshpelser:
**Affected version:** mutmut 3.6.0 (latest on PyPI); also present on `main` (`6453e6c`, 2026-06-10).

**Environment:** Python 3.14, Linux; reproduces independent of OS.

## Summary

During the stats/baseline phase, `record_trampoline_hit` (in
`src/mutmut/__main__.py`) unconditionally re-resolves the configured
`source_paths` with `Path.resolve(strict=True)` **on every trampoline hit**. The
resolved value is only ever used inside the `if Config.get().max_stack_depth != -1:`
block, and `max_stack_depth` defaults to `-1` — so in the default configuration
the resolve is pure dead work whose only effect is a failure mode.

`source_paths` are **relative** (e.g. `src/mypackage`). `resolve(strict=True)`
resolves them against the **current working directory**. Any test that `chdir`s —
for example, config tests that prove a `.env` file is loaded from the cwd — moves
the cwd somewhere that does not contain the relative `source_paths`, so the strict
resolve raises `FileNotFoundError` and aborts stats collection for the whole run.

## The offending code

`src/mutmut/__main__.py`, `record_trampoline_hit` (lines 121–135 on `main` as of
`6453e6c`; equivalent in 3.6.0):

```python
def record_trampoline_hit(name: str, caller: str | None = None) -> None:
    assert not name.startswith("src."), "Failed trampoline hit. Module name starts with `src.`, which is invalid"

    source_paths = [p.resolve(strict=True) for p in Config.get().source_paths]   # <-- runs on every hit, vs cwd

    if Config.get().max_stack_depth != -1:
        f = inspect.currentframe()
        c = Config.get().max_stack_depth
        while c and f:
            filename = f.f_code.co_filename
            f = f.f_back
            if "pytest" in filename or "hammett" in filename or "unittest" in filename:
                break
            file_path = Path(filename).resolve(strict=True)
            if any(path in file_path.parents for path in source_paths):
                c -= 1
        ...
```

The resolve happens before — and outside — the only guard (`max_stack_depth != -1`)
that consumes its result.

## Reproduction

1. Configure mutmut with relative `source_paths` (the normal case), leaving
   `max_stack_depth` at its `-1` default.
2. Include in the test suite any test that changes the working directory, e.g.:

   ```python
[...]

PR #527, opened 2026-06-10
Fix record_trampoline_hit when tmpdir changed working directory

Otto-AA:
Previously, if a test changed the working directory, then mutmut would fail to resolve the source_paths because it used relative paths to the new working directory.

As a fix, we now resolve the path before running the tests, thus with the original working directory.


Otto-AA:
The regression was added in https://github.com/boxed/mutmut/commit/2bf799a0f22c9021f71adec9f39bdc985e695447

Otto-AA:
Closes #526

PR #537, opened 2026-07-16
chore(readme): fix typo

scffs:
It uses MD link format in RST file but the «article» link above uses correct formatting style

https://github.com/boxed/mutmut/blob/main/README.rst?plain=1#L14


boxed:
Thanks!

Issue #541, opened 2026-07-29
Exception: Could not find mutant for type error


nedbat:
To reproduce:

```
% git clone https://github.com/nedbat/acidica
% cd acidica
% git checkout 3a142e473e319718edf42f514cbde304061d5b75
% uv venv -p 3.14
% source .venv/bin/activate
% uv sync
Resolved 29 packages in 8ms
warning: Skipping installation of entry points (`project.scripts`) for package `acidica` because this project is not packaged; to install entry points, set `tool.uv.package = true` or define a `build-system`
Prepared 4 packages in 217ms
Installed 27 packages in 77ms
 + ast-serialize==0.6.0
 + click==8.4.2
 + coverage==7.15.0
 + iniconfig==2.3.0
 + libcst==1.9.0
 + librt==0.13.0
 + linkify-it-py==2.1.0
 + markdown-it-py==4.2.0
 + mdit-py-plugins==0.6.1
 + mdurl==0.1.2
 + mutmut==3.6.0
 + mypy==2.3.0
 + mypy-extensions==1.1.0
 + packaging==26.2
 + pathspec==1.1.1
 + platformdirs==4.11.0
 + pluggy==1.6.0
 + pygments==2.20.0
 + pytest==9.1.1
 + pytest-cov==7.1.0
 + pyyaml==6.0.3
 + rich==15.0.0
 + setproctitle==1.3.7
 + textual==8.2.8
 + ty==0.0.58
 + typing-extensions==4.16.0
 + uc-micro-py==2.0.0

% uv pip install -e .
Resolved 1 package in 7ms
      Built acidica @ file:///private/tmp/acid2/acidica
Prepared 1 package in 792ms
Installed 1 package in 2ms
 + acidica==0.1.0 (from file:///private/tmp/acid2/acidica)

% pytest
====================================================================== test session starts =======================================================================
platform darwin -- Python 3.14.4, pytest-9.1.1, pluggy-1.6.0
rootdir: /private/tmp/acid2/acidica
configfile: pyproject.toml
plugins: cov-7.1.0
collected 140 items

tests/test_programs.py ................................................................................................................................... [ 93%]
.........                                                                                                                                                  [100%]

====================================================================== 140 passed in 0.12s =======================================================================

% mypy --strict src tests
Success: no issues found in 9 source files

% mutmut run
⠹ Generating mutants
    done in 6770ms (8 files mutated, 0 ignored, 0 unmodified)
⠙ Filtering mutations with type checker
[...]

PR #547, opened 2026-08-04
wip: don't mutate lines excluded from coverage

nedbat:
Written by Claude. I haven't reviewed it.


nedbat:
This now does what I needed it to do. I had code like this:
```
match something:
    ...

    case _NEVER:
        raise RuntimeError("Not implemented!")
```
and I have a coverage setting to exclude `case _NEVER:` clauses.  mutmut was mutating those cases by deleting them, and the tests wouldn't fail, because those cases never ran, as expected.

With this pull request, those cases are not mutated, because the coverage analysis is used to understand that they are excluded.

src/mutmut/mutation/file_mutation.py
```diff
@@ -619,6 +672,7 @@ def filter_mutants_with_type_checker() -> dict[str, FailedTypeCheckMutant]:
                     (m for m in mutated_methods if m.line_number_start <= error.line_number <= m.line_number_end), None
                 )
                 if mutant is None:
+                    continue
```
nedbat: This is here to prevent #541 from stopping the run.

boxed:
I looked through this a bit and it looks like a good improvement. The tests fail though, I guess that's because of the workaround for #541 ?

src/mutmut/mutation/file_mutation.py
nedbat: Oops, removed this.

boxed:
Great. This is merge-ready now then? Seems like an improvement over the code on main at least.

nedbat:
I don't know of a reason not to merge it. I really haven't read the code I'm sorry to say. I understand if you'd want to hold off.

boxed:
The only way is forward!

PR #549, opened 2026-08-08
Add mutation score badge support

rxdt:
Adds a mutation score badge users can put in their GitHub README to show the latest mutmut mutation score, as long as they have a JSON file the badge can read from.

Why this is useful:
- users get a simple public signal for test strength
- mutmut gets more visibility in READMEs
- teams can track score changes over time
- it brings mutmut closer to what Stryker already offers

Stryker does this with a hosted service: CI uploads the score, and the badge reads it back from Stryker. Mutmut does not have a server, so the right place for this code is in mutmut itself. It generates the badge JSON, and the user can host that file wherever they want.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -518,41 +518,18 @@ this directory to start from scratch.
 Mutation score badges
 ---------------------
 
-If you want a mutation score badge in GitHub, mutmut can turn its CI stats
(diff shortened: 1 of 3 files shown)


boxed:
Thanks!

Issue #373, opened 2025-03-24
How to specify runner, create JUnit XML in mutmut 3?


l0b0:
Here are [my original mutmut CI commands](https://gitlab.com/engmark/mypy-exercises/-/blob/38e5d1ddaedf890993e1705db8e4b3ff92252a45/.gitlab/mutation-test.bash):

```bash
mutmut run --no-progress --runner="pytest --assert=plain --exitfirst --randomly-seed=${CI_JOB_ID:?}" ${@+"$@"}"
mutmut junitxml > mutmut.xml
```

But `--no-progress` (which is not crucial), `--runner`, and `junitxml` are all gone from mutmut 3, and there doesn't seem to be configuration options to achieve the same thing. [Another comment mentioned stronger integration with pytest](https://github.com/boxed/mutmut/issues/368#issuecomment-2695290579), but I don't see that mentioned in the [docs](https://mutmut.readthedocs.io/en/latest/). How can I achieve the same thing with mutmut 3?

Also, a plain `mutmut run` hits [this assertion](https://github.com/boxed/mutmut/blob/b6fc4bc374a32d590b1b71b56f99720ad4027b1c/mutmut/__main__.py#L139), which I don't understand. Why treat the name `src` specially? And what's the fix? Just rename the directory anything else?

l0b0:
I've tried to configure mutmut, but I keep getting weird issues. With the [latest state](https://gitlab.com/engmark/mypy-exercises/-/commit/ce4b0fce2e3c48fb5e59512ef6813381403f4941), pytest succeeds, but mutmut fails to run the tests afterwards. Then running pytest after mutmut reveals that the `__pycache__` files from mutmut has broken pytest:

<details>
<summary>Details</summary>
```
❯ pytest
============================================================================= test session starts ==============================================================================
platform linux -- Python 3.12.9, pytest-8.3.3, pluggy-1.5.0
Using --randomly-seed=1329102626
rootdir: /home/victor/my projects/mypy-exercises
configfile: pyproject.toml
plugins: subtests-0.13.1, randomly-3.15.0
collected 38 items                                                                                                                                                             

source/annotated/test_all.py ,,,,.,,,,.,,.,,,,.,,,,,,.,,.                                                                                                                [ 15%]
source/builtin/test_all.py ,,.                                                                                                                                           [ 18%]
source/generic/test_all.py ,,.                                                                                                                                           [ 21%]
[...]

PR #550, opened 2026-08-11
fix: support legitimate top-level src packages

saagpatel:
## Summary

- recognize `src/__init__.py` as evidence that `src` is the actual top-level package
- preserve `src.` consistently in trampoline hits, mutant keys, hashes, and type-checker filtering
- keep the existing guard and prefix stripping for conventional src-layout projects
- migrate stripped keys left by a failed pre-fix run without requiring manual cache deletion

This addresses the top-level `src` package failure described in #373 while preserving the safeguard discussed by the maintainer in that thread.

## Root cause and reproduction

A project with `src/__init__.py`, tests importing `src.calculator`, and `source_paths = ["src/"]` passes pytest but fails mutmut during stats collection:

```text
AssertionError: Failed trampoline hit. Module name starts with `src.`, which is invalid
```

The assertion and path-derived key normalization made conflicting assumptions: Python records `src.calculator...`, while mutmut always strips the `src.` prefix from the corresponding mutant key. Relaxing only the assertion would therefore leave stats and mutant keys mismatched.

## Behavior and compatibility

Mutmut now detects the package marker during configuration loading and preserves `src.` only for that case. Projects using a conventional src layout without `src/__init__.py` retain the existing fail-fast behavior and stripped keys.

For an affected project with a cache generated before this fix, mutmut detects legacy stripped keys, regenerates the file once, and then resumes the existing unmodified-file cache path. A direct regression test covers that migration.

The optional `preserve_src_prefix` argument is keyword-only and defaults to the established behavior. `Config.src_package_exists` also has a compatibility-preserving default. No dependency, lockfile, command-line, or serialized-format change is included.

## Verification

- new end-to-end fixture for an actual top-level package named `src`
- 7 focused guard, configuration, naming, cache-migration, and end-to-end tests passed on Python 3.10.20 and 3.14.6
- full local suite: 386 passed, 2 skipped
- all pre-commit hooks passed, including Ruff and mypy
- current consumer `saagpatel/GithubRepoAuditor@3c0e369`: 155 focused tests passed, and all four selected `src.scorer.x_letter_grade` mutants were killed with exit status 0
- baseline `boxed/mutmut@5cb006d` reproduces the consumer failure at `record_trampoline_hit`

## Scope and risk

The detection intentionally requires `src/__init__.py`; PEP 420 namespace packages named `src` remain out of scope. Conventional src layouts are unchanged. The patch adds one project-local `is_file()` check during configuration loading; only an affected legacy cache pays a one-time regeneration cost.

This draft does not address the separate runner or JUnit requests in #373.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,6 +99,11 @@ If you use `pyproject.toml`, you must specify the paths as array in a `tool.mutm
     source_paths = [ "src/" ]
     pytest_add_cli_args_test_selection= [ "tests/" ]
 
+For a standard ``src`` layout, import the package inside ``src`` rather than
+importing ``src`` itself. If ``src/__init__.py`` exists because ``src`` is the
+actual top-level package, mutmut recognizes that package and preserves its
+``src.`` module prefix in mutant names.
+
 See below for more options for configuring mutmut.
 
 
--- a/e2e_projects/src_package/pyproject.toml
+++ b/e2e_projects/src_package/pyproject.toml
@@ -1,0 +1,6 @@
+[tool.mutmut]
+source_paths = ["src/"]
+pytest_add_cli_args_test_selection = ["tests/"]
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
--- a/e2e_projects/src_package/src/__init__.py
+++ b/e2e_projects/src_package/src/__init__.py
@@ -1,0 +1,1 @@
+"""A legitimate top-level package named src."""
--- a/e2e_projects/src_package/src/calculator.py
+++ b/e2e_projects/src_package/src/calculator.py
@@ -1,0 +1,2 @@
+def add_one(value: int) -> int:
+    return value + 1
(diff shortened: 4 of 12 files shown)


boxed:
I don't get it. This seems like adding support for broken projects. I'd rather mutmut tell them to fix their stuff.

saagpatel:
Fair point — I probably didn’t explain the distinction well.

This isn’t a normal src-layout project accidentally importing from src. Here, src is intentionally the package name: it has an __init__.py, setuptools ships src*, and the installed entry points import src.*. It’s definitely unconventional, but it isn’t an import-path mistake.

That said, I don’t want mutmut to silently allow broken src layouts either. Would you be open to making this an explicit opt-in instead of auto-detecting it? The current fail-fast behavior would remain the default, and a project would have to deliberately say, “yes, my package is actually named src.”

If you’d rather mutmut not support that layout at all, that’s fair too — I can close this and keep the workaround downstream.

boxed:
This is just not valid. Just look at the entry for src on pypi: https://pypi.org/project/src/

Mutation testing is about increasing the quality of your project. If your project has this layout you should fix this WAY before doing mutation testing 😅

Issue #544, opened 2026-07-31
run_stats records the teardown duration per test (missing call.when filter), starving the per-mutant timeout budget


JonasPammerWork:
# [boxed/mutmut] run_stats records the TEARDOWN duration per test (missing `call.when` filter) → near-zero estimates starve the per-mutant budget

## Summary

The stats-collection plugin's `pytest_runtest_makereport` hook stores `call.duration` **without filtering on `call.when`**. The hook fires three times per test (setup / call / teardown) and the storage is last-write-wins, so the recorded per-test duration is the **teardown** figure — typically well under 10 ms regardless of how expensive the actual test call was.

Measured on our suite (`mutants/mutmut-stats.json`): 210 of 219 tests recorded under 0.01 s, median 0.0039 s — while the real wall time of the covering set for our hottest mutation targets is ~25–31 s.

## Why it matters

`timeout_checker` derives the per-mutant budget from these stats:

- wall deadline: `(estimated_time_of_tests + 1) * 15`
- `RLIMIT_CPU`: `(estimated_time_of_tests + 1) * 30`

With `estimated_time_of_tests ≈ 0`, every mutant gets a ~15 s wall budget **no matter what the covering tests really cost**. Mutants whose covering set genuinely needs more are scored ⏰ (timeout) instead of 🎉/🙁 — and a timed-out mutant is not scored at all, so heavy, well-tested code paths systematically drop out of the mutation score. No suite-level speedup can compensate, because the estimate never reflects the call phase in the first place.

On our project this clipped 366–418 mutants per run as spurious timeouts until we worked around it by shrinking the covering set's real cost (hypothesis `max_examples` gating on `MUTANT_UNDER_TEST`).

## Suggested fix

In the makereport hook, record only the call phase (or sum the phases):

```python
def pytest_runtest_makereport(self, item, call):
    if call.when != "call":
        return
    ...
```

## Environment

- mutmut 3.x (config-driven `[tool.mutmut]`, `paths_to_mutate`)
- pytest 8.x, Django project, suite driven via `mutmut run`
- Related but distinct: #518 (est-lookup uses the wrong key); this report is about the recorded value itself.

PR #510, opened 2026-04-27
[TMP] Feature Preview

nicklafleur:
This PR serves as a public preview of our in-flight work

PR #561, opened 2026-09-03
refac: main decomposition

nicklafleur:
3 more or less independent commits that work towards the same ultimate goal.

### 1. `refactor(config)`: unify `Config`/`MutmutState` to function-based singletons

`Config` still exposed `Config.get()` / `ensure_loaded()` / `reset()` classmethods while
`MutmutState` (#509) already used a bare `state()` accessor. Replaces the classmethod API
with module-level `config()` / `reset_config()` to match, and migrates the last
module-level globals in `mutmut/__init__.py` (`stats_time`, `duration_by_test`,
`tests_by_mangled_function_name`, `_stats`, `_covered_lines`, `_excluded_lines`) onto
`MutmutState`. Deprecated `__getattr__` shims are kept for external readers of the old
`mutmut.*` attributes.

Test fixtures that patched `Config.get()` at the class level now mutate the live singleton
in place, so patches stay visible to modules that imported `config` by reference.

### 2. `refactor(workers)`: create workers package

Renames `mutmut/threading/` to `mutmut/workers/` (and the matching test package). Pure
rename; `timeout.py` is unchanged.

### 3. `refactor`: extract `__main__.py` components into dedicated modules

Moves cohesive groups out of the oversized `__main__.py` (2064 → 1399 lines):

- `stats.py`: status/emoji maps, `Stat`, `collect_stat`, `calculate_summary_stats`, `print_stats`, `load_stats`, `save_stats`
- `runners/harness.py`: `TestRunner` ABC, `PytestRunner`, `HammettRunner`, `ListAllTestsResult`, test-runner exceptions
- `ui/browse.py`: `ResultBrowser` wrapped in `run_result_browser()`; `get_diff_for_mutant`/`apply_mutant` are injected to avoid a circular import back into `__main__`
- `ui/terminal.py`: spinner + `status_printer`/`print_status`
- `utils/file_utils.py`: the `walk_*` helpers, `copy_src_dir`, `copy_also_copy_files`, `setup_source_paths`

`result_browser_layout.tcss` moves into `ui/` so Textual's `CSS_PATH` resolves relative to
`ui/browse.py`. `__main__.py` re-imports the moved public names, so external
`from mutmut.__main__ import ...` call sites keep working unchanged.

### Notes for reviewers

These commits were written before several fixes landed on `main`, so I needed to re-apply some changes,
would appreciate a review on that.

- the `call.when != "call"` guard (#544) now lives in `runners/harness.py`
- `_excluded_lines` (#547) was added to `MutmutState`, since it postdates the config refactor


Otto-AA:
Looks good to me.

Some minor things I noticed (feel free to address, not necessary):
- I think passing `get_diff_for_mutant` and `apply_mutant` as arguments to `run_result_browser` makes it harder to navigate from the usage within `run_result_browser` to the actual definition of these functions. I'd tend to call them directly. You could also move them to a some other file that __main__.py and browse.py both import
- the `if TYPE_CHECKING: pass` in __main__.py is unnecessary

And do you plan to put more code into the workers module? Currently, it feels a bit empty. I guess the `fork` stuff will/could be extracted from __main__.py and moved into the workers module?

Otto-AA (approved):

nicklafleur:
> Looks good to me.
> 
> Some minor things I noticed (feel free to address, not necessary):
> 
> * I think passing `get_diff_for_mutant` and `apply_mutant` as arguments to `run_result_browser` makes it harder to navigate from the usage within `run_result_browser` to the actual definition of these functions. I'd tend to call them directly. You could also move them to a some other file that **main**.py and browse.py both import
> * the `if TYPE_CHECKING: pass` in **main**.py is unnecessary
> 
> And do you plan to put more code into the workers module? Currently, it feels a bit empty. I guess the `fork` stuff will/could be extracted from **main**.py and moved into the workers module?

I have more coming for the workers package (namely forking strategies you can find in #510). I got a little stuck since I changed jobs and lost access to the computer I was doing this work with, but I'm set back up now so will be merging it all in short order.

Fixed the issues pointed out in your review, thanks

Issue #528, opened 2026-06-12
mutate_only_covered_lines crashes on C-extension deps (numpy, libyaml): cannot load module more than once per process


ejfine:
### Summary

  When `mutate_only_covered_lines = true`, the coverage-gathering phase unloads
  **every** module imported during its test run (`_unload_modules_not_in`), including
  third-party C extensions. The subsequent stats run then re-imports them in the same
  process. C extensions that forbid re-initialization crash:

  - `numpy` → `ImportError: cannot load module more than once per process`
  - `PyYAML` C loader (`_yaml`, used by VCR/vcrpy cassettes) → corrupted parse:
    `yaml.constructor.ConstructorError: could not determine a constructor for the tag None`

  **This is not test-specific.** The crash depends only on *when* the C extension first
  loads, not on who imports it. It is triggered by any single-init C extension that is
  first imported during the coverage run (after mutmut's `sys.modules` snapshot) and then
  re-imported during the stats run — whether that import originates in a test module or in
  the source code under mutation. A `source_paths` module that imports numpy crashes the
  same way as a test that imports it.

  Setting `mutate_only_covered_lines = false` avoids it entirely (no coverage pre-run →
  the deps import exactly once in the stats run; per-mutant runs `os.fork()` and inherit
  via copy-on-write), which is what points at the unload as the cause.

  ### Environment

  - mutmut 3.6.0
  - Python 3.13.9 (Linux)
  - pytest 9.x, numpy, vcrpy + PyYAML built with libyaml

  ### Reproduction

  `pyproject.toml`:
  ```toml
  [tool.mutmut]
  source_paths = ["src/mypkg/"]
  pytest_add_cli_args_test_selection = ["tests/unit/"]
  mutate_only_covered_lines = true
```

  A single unit test importing numpy is enough:
```
  from numpy.testing import assert_almost_equal

  def test_close():
      assert_almost_equal(1.0, 1.0)
```

  Equivalently, importing numpy from a src/mypkg/ module under mutation (with any test
  that exercises it) reproduces the identical crash — confirming the issue is not confined
  to the test suite.

  mutmut run fails in the stats phase:
  ImportError: cannot load module more than once per process
  (For libyaml, any vcrpy cassette test produces the tag None ConstructorError instead.)

  Root cause

  mutmut/code_coverage.py gather_coverage():
  modules = dict(sys.modules)          # snapshot before the coverage run
  cov = coverage.Coverage(data_file=None)
[...]

Issue #474, opened 2026-02-28
Support parallel tests with external database


Otto-AA:
Similar to pytest-xdist: https://pytest-xdist.readthedocs.io/en/stable/how-to.html

I'm not sure yet how to implement this in mutmut, as the pytest setup is different (in pytest-xdist, session fixtures are run once per worker. Not sure how we could replicate this use case in mutmut)

I have a project which connects to a local postgres instance for integration tests. When running tests in parallel, they should connect to different databases on this instance, otherwise they would conflict with each other (e.g. test A deletes data from some table, while test B tries to read from this table). This means I currently cannot run these tests with mutmut.

The setup with pytest-xdist is:
1. start e.g. 12 processes with `PYTEST_XDIST_WORKER` set to `gw1`, `gw2`, ..., gw12`
2. for each process, it runs the pytest session fixture to setup the database
2.1. The fixture connects to postgres and creates a database `test_db_{PYTEST_XDIST_WORKER}`
2.2. Afterwards, we have 12 different databases
3. When running a test, it runs against the `test_db_{PYTEST_XDIST_WORKER}` database

Therefore, each worker uses its own database.

I think something like step (1) would be rather straightforward: We use `fork` and do not have explicit workers, but we run at most `n` processes in parallel. We could create `n` ids, keep track of which ids are still free, and everytime we `fork` another process we take one of the free ids.

However, I'm not sure how we would go about running session fixtures once per "worker". I suppose, we would could:
1. `spawn` (and not `fork`) 12 worker processes
2. set `MUTMUT_WORKER_1/2/.../12` 
3. In each process run the complete test suite / an init method (for caching and running the database setup fixtures)
4. Split the mutants across the workers (or use a multiprocessing queue/...)
5. Inside of the worker, we sequentially run a single `fork` to create isolated copies of the worker process (with everything already setup) and run tests for mutations there
6. Workers write their results into a results queue and the main process saves the stats in files (we should not write the `foo.py.meta` files from different processes)

nicklafleur:
I have some new forking strategies I needed to implement to support some libraries we use at Lyft (gevent) coming up in a PR (I believe it's either next or after 1 re-org PR, need to check) which partially addresses this kind of thing. I had initially moved away from the subprocess implementation as it resulted is slower performance, but the xdist-like use-case was not one I had considered when removing it. We could fairly easily add a new run mode which can better handle this kind of case by extending the new abstraction (MutantRunner) to handle this behaviour.

[...]

PR #566, opened 2026-09-07
feat: process isolation and fork-safe test runners

nicklafleur:
## Pluggable process isolation, with a fork-safe `hot-fork` runner

Adds a `MutantRunner` abstraction that owns the process-isolation strategy for mutation testing

Two strategies:

- **`fork`** (default) the existing fork-per-mutant behaviour, extracted into `ForkRunner`. 
  No behavioural change.
- **`hot-fork`** `parent (clean) -> orchestrator (imports pytest once) -> N grandchildren`,
  for projects whose test setup makes the parent fork-unsafe (gevent monkey-patching,
  grpc, torch). The parent never imports pytest or conftest; the orchestrator imports it
  exactly once and forks a grandchild per mutant. Results stream back over a
  length-prefixed pipe protocol, and grandchildren are reaped via a SIGCHLD self-pipe.
  A crashed orchestrator is restarted with its in-flight mutants re-submitted, up to
  `max_orchestrator_restarts`, before raising `OrchestratorCrashError`.

Stats collection, clean tests, forced-fail and test listing all run in short-lived forks
(`run_in_fork` / `run_in_fork_with_result`) so the parent stays clean throughout.

### Config

| Option | Default | Purpose |
| --- | --- | --- |
| `process_isolation` | `fork` | `fork` or `hot-fork` |
| `hot_fork_warmup` | `collect` | `collect` / `import` / `none` |
| `max_orchestrator_restarts` | `3` | Restarts before `OrchestratorCrashError` |
| `preload_modules_file` | `None` | Modules to preload in the orchestrator |
| `log_to_file` | `False` | Enable orchestrator file logging |
| `log_file_path` | `mutants/mutmut-debug.log` | Log destination |

### Incidental cleanups

Breaks the `mutation/trampoline.py` -> `__main__` import cycle: `MutmutProgrammaticFailException`
moves to `core.py` and `record_trampoline_hit` to `stats.py`, which removes the
`_set_mutant_under_test` late-import shim. `src/` now has zero `from mutmut.__main__`
imports. Also drops the dead pid-keyed bookkeeping from `SourceFileMutationData`
(`register_pid` / `register_result` / `stop_children`), superseded by `MutantResult`, and
sets `platform = linux` in `mypy.ini`, mutmut is (currently) POSIX-only, so type-checking 
elsewhere otherwise flags `os.fork` / `resource.setrlimit` / `signal.SIGXCPU` as missing.

### Testing

Verified that `fork` and `hot-fork` produce identical verdicts on the `my_lib` e2e project.
New coverage for the pipe protocol (including payloads larger than the pipe buffer),
`ForkRunner`/`HotForkRunner` no-test-mutant handling, `get_mutant_runner` dispatch,
`OrchestratorCrashError` messaging, and config validation, plus a subprocess e2e smoke
test (`tests/e2e/test_e2e_hot_fork.py` with the `hot_fork_basic` project) exercising the
full hot-fork path.


Otto-AA:
Thanks for the PR :heart: 

I'll try to look a bit more at it today or tomorrow.

From a high level, I think we get the following (?):

| Does it ...  | fork | hot-fork |
| -- | -- | -- |
| Cache all imports? (pytest and mutated source code) | Yes (we ran the whole pytest suite) | Yes (configurable via `hot_fork_warmup`, either `pytest --collect-only` or fixed list of imports or nothing) |
| Cache session fixtures | Not sure actually, whether executing pytest twice from python code will call the session fixture twice. | No |
| Run pytest multiple times within the same process? | Yes | No (except with `mutate_only_covered_lines` ?)

I think the hot-fork + mutate_only_covered_lines combination is mabye not as expected: We run pytest via `gather_coverage` first in the main process and later on fork the orchestrator with pytest already being loaded. Should we also run the coverage collection via the 

What this PR would fix:
- Fixes #578 (with the hot-fork option, because nothing "process-global" is created before forking)

What this PR nearly fixes:
- Fixes #528 (with the hot-fork option. But I think because we run coverage collection in the main process and then fork afterwards in startup, we still have the same problem)

What this PR does not attempt to fix:
- #474 (at least in my use case, we use session fixtures to setup database tables and connections. I think session fixtures are not usable in the hot-fork scenario)

tests/e2e/test_e2e_hot_fork.py
```diff
@@ -0,0 +1,54 @@
...
+def _read_verdicts(project_path: Path) -> dict[str, int]:
+    verdicts: dict[str, int] = {}
+    for meta in sorted((project_path / "mutants").rglob("*.meta")):
+        data = json.loads(meta.read_text())
+        verdicts.update(data["exit_code_by_key"])
+    return verdicts
+
+
+def test_hot_fork_runs_end_to_end(tmp_path: Path):
+    project = tmp_path / "hot_fork_basic"
```
Otto-AA, line 30: Is there a reason you use a different setup than for the other tests?

This seems to work:

```python
def test_hot_fork_runs_end_to_end():
    assert run_mutmut_on_project("hot_fork_basic") == snapshot(
        {
            "mutants/src/hf_calc/__init__.py.meta": {
                "hf_calc.x_add__mutmut_1": 1,
                "hf_calc.x_sub__mutmut_1": 1,
                "hf_calc.x_mul__mutmut_1": 1,
                "hf_calc.x_untested__mutmut_1": 33,
                "hf_calc.x_untested__mutmut_2": 33,
            }
        }
    )
```

Otto-AA (approved):
I won't have time to review the code in the next weeks, so here are some high level inputs (feel free to incorporate or to merge without changes):

- the `process_isolation` and `hot_fork_warmup` should be explained in the README. As a user, I'd like to know when to use which of those.
- I feel like the naming `fork` and `hot-fork` does not reflect the differences between those two. I'm not sure how to better name them though
- I feel like it could fix #528, but currently does not. I haven't verified it though. (could also be fixed in a separate PR)

tests/e2e/test_e2e_hot_fork.py
nicklafleur: I believe this branch predated adding the snapshot to the project and I must have just gone blind to it. nice catch!

nicklafleur:
> I won't have time to review the code in the next weeks, so here are some high level inputs (feel free to incorporate or to merge without changes):
> 
> * the `process_isolation` and `hot_fork_warmup` should be explained in the README. As a user, I'd like to know when to use which of those.
> * I feel like the naming `fork` and `hot-fork` does not reflect the differences between those two. I'm not sure how to better name them though
> * I feel like it could fix #528, but currently does not. I haven't verified it though. (could also be fixed in a separate PR)

good call on these suggestions, especially 528. I'll see if I can slot these in neatly. I agree hot-fork is not the best name, I named it based on the idea that the fork gets warmed, but I should find something better. I'll see what claude can come up with, definitely has better ideas than I can come up with for naming things.

PR #568, opened 2026-09-08
Fix clear_out_obsolete_test_names never persisting its cleanup

### LINKED ISSUE #564: clear_out_obsolete_test_names never persists its cleanup: counters iterate keys instead of values
`ListAllTestsResult.clear_out_obsolete_test_names` (`src/mutmut/__main__.py`, lines 421–433 on `main`) filters obsolete test names out of `tests_by_mangled_function_name` correctly, but the guard around `save_stats()` can never be true, so the cleaned state is never written back to `mutants/mutmut-stats.json`.

```python
count_before = sum(len(x) for x in mutmut.tests_by_mangled_function_name)   # line 422
...
count_after  = sum(len(x) for x in mutmut.tests_by_mangled_function_name)   # line 430
if count_before != count_after:
    print(f"Removed {count_before - count_after} obsolete test names")
    save_stats()
```

`tests_by_mangled_function_name` is a `defaultdict(set)` (`src/mutmut/__init__.py`, line 15). Iterating it yields **keys**, so `len(x)` measures the length of each mangled function name, not the size of its test set. The dict comprehension in between preserves every key, so both sums are computed over an identical key set and `count_before == count_after` **always** — not intermittently. The branch is dead: the message never prints and `save_stats()` never runs from here.

### Impact

Limited but persistent. The in-memory filtering happens before `tests_for_mutant_names` is used, so runs are unaffected and no verdict is wrong. What lingers is the file on disk: obsolete node IDs stay in `mutmut-stats.json` indefinitely, unless something else happens to call `save_stats()` in the same process (for example `run_stats_collection` when new tests are found). Anyone reading the stats file therefore sees a state no run has ever used.

### Reproduce

1. Run `mutmut run` on a project until stats are cached.
2. Delete or move one test file.
3. Run `mutmut run` again.
4. `mutants/mutmut-stats.json` still contains that file's node IDs under `duration_by_test` and `tests_by_mangled_function_name`, and `Removed N obsolete test names` is never printed.

### Suggested fix

```python
count_before = sum(len(x) for x in mutmut.tests_by_mangled_function_name.values())
count_after  = sum(len(x) for x in mutmut.tests_by_mangled_function_name.values())
```

in both places, which also makes the reported number mean what the message says.

Observed on 3.7.0; the code is unchanged on `main`. Happy to send a PR if that is easier than fixing it directly.

t-roi33:
Fixes #564.

`count_before` and `count_after` in `ListAllTestsResult.clear_out_obsolete_test_names` iterated the keys of `tests_by_mangled_function_name`, so they summed the lengths of the mangled names rather than the sizes of the test sets. Since the comprehension in between keeps every key, the two sums were always equal, the "Removed N obsolete test names" message never printed and `save_stats()` never ran from here. The in-memory filtering worked, but obsolete node IDs stayed in `mutants/mutmut-stats.json` forever.

This counts the test sets instead and adds a unit test that drives the real `save_stats()` into a tmp dir and checks the cleanup reaches the file (plus one for the no-op case). The code moved from `__main__.py` to `runners/harness.py` in #561, but is otherwise unchanged from the issue.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,8 @@ Changelog
 Unreleased
 ~~~~~~~~~~
 
+* Fix obsolete test names never being removed from ``mutants/mutmut-stats.json``: the counters guarding the save iterated dict keys instead of the test sets, so the cleanup was applied in memory but never persisted (`#564`)
+
 * Fix ``# pragma: no mutate block`` being silently ignored when placed on an ``else``, ``except``, ``except*`` or ``finally`` header, and on the ``try`` line of a ``try``/``except*``
 
 * Fix mutants being reported as survived when a test uses ``patch.dict(os.environ, ..., clear=True)`` (`#511`)
--- a/src/mutmut/runners/harness.py
+++ b/src/mutmut/runners/harness.py
@@ -60,7 +60,7 @@ class ListAllTestsResult:
         self.ids = ids
 
     def clear_out_obsolete_test_names(self) -> None:
-        count_before = sum(len(x) for x in state().tests_by_mangled_function_name)
+        count_before = sum(len(x) for x in state().tests_by_mangled_function_name.values())
         state().tests_by_mangled_function_name = defaultdict(
             set,
             **{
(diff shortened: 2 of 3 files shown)


nicklafleur:
we'll merge with #569

PR #569, opened 2026-09-08
Fix obsolete test names never being removed from the stats file

### LINKED ISSUE #564: clear_out_obsolete_test_names never persists its cleanup: counters iterate keys instead of values

t-roi33:
Fixes #564. Replaces #568, whose commit is the first of the two here.

`clear_out_obsolete_test_names` is meant to drop the node IDs of deleted or
renamed tests from `mutants/mutmut-stats.json`. It never did, for two reasons:

* `count_before` and `count_after` iterated the keys of
  `tests_by_mangled_function_name`, so they summed the lengths of the mangled
  names rather than the sizes of the test sets. The comprehension in between
  keeps every key, so the two sums were always equal, the "Removed N obsolete
  test names" message never printed and `save_stats()` never ran.
* The filtering itself only covered `tests_by_mangled_function_name`. Obsolete
  node IDs in `duration_by_test` were never touched and stayed in the file for
  good.

Both come off one computation now: obsolete names are everything the stats file
knows minus what the collection run found. Both structures are pruned with that
set, and its size is what the message reports, which is what it says.

`new_tests()` is unaffected: it only looks at what is in the collection result
but not in the stats, and this removes nothing from the former.

The tests drive the real `save_stats()` into a tmp dir and check that both
structures reach the file pruned, plus the no-op case.


nicklafleur (approved):
nice, I closed the other PR, we'll merge just this, in the future feel free to just combine the fixes for something like this.

Just rework the title and message a bit to reflect both fixes, thanks

HISTORY.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 Unreleased
 ~~~~~~~~~~
 
+* Fix obsolete test names never being removed from ``mutants/mutmut-stats.json``: the counters guarding the save iterated dict keys instead of the test sets, so the cleanup was applied in memory but never persisted (`#564`)
+
+* Also drop obsolete test names from ``duration_by_test`` in ``mutants/mutmut-stats.json``, not only from the test-to-function mapping
```
nicklafleur:
```suggestion
* Fix obsolete test names never being removed from ``mutants/mutmut-stats.json`` (`#564`)
```

PR #579, opened 2026-09-10
Add configurable mutation type filtering

### LINKED ISSUE #577: String mutations: the two case flips are equivalent wherever the consumer is case-insensitive, and no option turns them off
> **Disclaimer:** AI-drafted, verified by me before filing. I answer follow-ups.

`operator_string` emits three mutants for every string literal: the literal wrapped in `XX`, lowercased, and uppercased. Where the literal's consumer is case-insensitive -- SQL, codec names, HTTP header names -- the case flips are equivalent mutants by construction.

**Expected:** a mutation I cannot kill can be switched off, as `--disable-mutation-types` did in 2.x (#175, added by #225).
**Actual:** both case flips survive every run, and the only lever in 3.x, `do_not_mutate_patterns`, matches whole lines, so it drops the `XX` mutant on that line too.

They were 25% and 11% of all survivors on two of my codebases: re-read on every run, and nothing to be done about them.

<details>
<summary><b>Reproduce</b> -- three files, mutmut 3.7.0, pytest 9.1.1, CPython 3.12.12</summary>

```python
# example.py
import sqlite3


def note_count(db_path: str) -> int:
    con = sqlite3.connect(db_path)
    return con.execute("SELECT COUNT(*) FROM notes").fetchone()[0]
```

```python
# test_example.py
import sqlite3

from example import note_count


def test_note_count(tmp_path):
    db = tmp_path / "n.db"
    con = sqlite3.connect(db)
    con.execute("CREATE TABLE notes (id int)")
    con.execute("INSERT INTO notes VALUES (1)")
    con.commit()
    con.close()
    assert note_count(str(db)) == 1
```

```toml
# pyproject.toml
[project]
name = "repro"
version = "0"
requires-python = ">=3.11"

[tool.mutmut]
source_paths = ["example.py"]
```

```console
$ mutmut run
7/7  🎉 5 🫥 0  ⏰ 0  🤔 0  🙁 2  🔇 0  🧙 0

$ mutmut results
    example.x_note_count__mutmut_5: survived
    example.x_note_count__mutmut_6: survived

$ mutmut show example.x_note_count__mutmut_5
# example.x_note_count__mutmut_5: survived
--- example.py
+++ example.py
@@ -1,3 +1,3 @@
 def note_count(db_path: str) -> int:
     con = sqlite3.connect(db_path)
-    return con.execute("SELECT COUNT(*) FROM notes").fetchone()[0]
+    return con.execute("select count(*) from notes").fetchone()[0]

$ mutmut show example.x_note_count__mutmut_6      # the same line, uppercased
+    return con.execute("SELECT COUNT(*) FROM NOTES").fetchone()[0]
```

SQLite parses either casing and resolves table names case-insensitively, so both mutants are equivalent to the original. The only test that would kill them asserts the query string's exact casing, which tests the literal rather than the behaviour.

</details>

<details>
<summary><b>What the line pattern costs</b> -- 7 mutants become 2</summary>

```toml
[tool.mutmut]
source_paths = ["example.py"]
do_not_mutate_patterns = ["SELECT"]
```

```console
$ mutmut run
2/2  🎉 2 🫥 0  ⏰ 0  🤔 0  🙁 0  🔇 0  🧙 0
```

Five of the seven mutants are gone. Two of them, mutant 4 (`"XXSELECT COUNT(*) FROM notesXX"`) and mutant 7 (`.fetchone()[1]`), are killed by the test as it stands, so the pattern removes working signal in order to remove two mutants that carry none.

</details>

nabi-noor:
Adds stable dotted names to the mutation registry and a `disable_mutation_types` configuration option. Exact names disable one mutation, while prefixes disable a group; for example, `string.lower` removes only lowercase literal mutations and `string` removes every string mutation.

The filtering happens before an operator runs. Configuration changes also invalidate cached mutant verdicts because removing a mutation can shift the numeric IDs of later mutants in the same function.

This PR intentionally focuses on disabling types, which solves the reported equivalent-mutant problem without defining potentially surprising precedence between simultaneous allow and deny lists. An enable/override policy can be added separately once those semantics are agreed.

Documentation lists the available names and shows how to disable the two case-flip mutations. Unit coverage exercises exact, grouped, unrelated, invalid, and cache-change behavior; the existing config E2E project verifies that both case flips disappear while the useful string-wrap mutant remains.

Verified with `uv run pytest` (403 passed, 2 skipped) and `uv run pre-commit run --all-files` (all hooks passed).

AI assistance disclosure: I used OpenAI Codex while implementing and reviewing this change. The commit includes the corresponding co-author attribution.

Closes #577

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -246,6 +246,41 @@ via the following options. Examples where this could be relevant:
 - Optimizing break instead of continue. The code runs fine when mutating break
   to continue, but it's slower.
 
+Skipping mutation types
+^^^^^^^^^^^^^^^^^^^^^^^
+
+You can disable mutation types that are not useful for your codebase. For example,
+SQL keywords and HTTP header names are often case-insensitive, so their lowercase
+and uppercase string mutants may be equivalent:
+
+.. code-block:: toml
+
+    # pyproject.toml
+    [tool.mutmut]
+    disable_mutation_types = ["string.lower", "string.upper"]
+
+Mutation types use dotted groups. A group name disables every type below it, so
+``disable_mutation_types = ["string"]`` disables all string mutations.
+
+The available mutation types are:
+
+- ``argument.keyword``
+- ``argument.removal``
+- ``assignment``
+- ``assignment.augmented``
+- ``if_expression``
+- ``keyword``
+- ``lambda``
+- ``match``
+- ``name``
+- ``number``
+- ``operator``
+- ``operator.unary``
+- ``string.lower``
+- ``string.method.swap``
+- ``string.upper``
+- ``string.wrap``
+
 Skipping via regex
 ^^^^^^^^^^^^^^^^^^
 
--- a/e2e_projects/config/pyproject.toml
+++ b/e2e_projects/config/pyproject.toml
@@ -27,6 +27,7 @@ debug = true
 source_paths = [ "config_pkg/" ]
 only_mutate = [ "config_pkg/logic/*" ]
 do_not_mutate = [ "*ignore*" ]
+disable_mutation_types = [ "string.lower", "string.upper" ]
 also_copy = [ "data" ]
 max_stack_depth=3
 # verify that we can override options with pytest_add_cli_args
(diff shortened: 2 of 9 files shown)


nicklafleur:
@nabi-noor mind rebasing and resolving conflicts?

PR #580, opened 2026-09-10
Mutate f-strings and concatenated strings

### LINKED ISSUE #439: Mutate f-strings and concatenated strings
Our `operator_string` currently only mutates `cst.SimpleString`, but not `cst.FormattedString` and `cst.ConcatenatedString`.

I think we should also mutate those, similar to the simple string. I would only mutate one (the first non-empty?) part of the string parts. The other elements can be kept the same, because I think mutating one location per string is enough to check if it is covered by a test, and leaving the others out we can reduce the number of mutants.

nabi-noor:
## Summary

- apply the existing string mutations to literal text in formatted strings
- treat implicitly concatenated strings as one unit and mutate only the first non-empty text segment
- avoid duplicate child-string mutants while preserving mutations inside f-string expressions
- add focused regression cases and update the affected end-to-end snapshots

Closes #439

## Testing

- `uv run pytest tests/mutation/test_mutation.py tests/e2e/test_e2e_coverage.py tests/e2e/test_e2e_my_lib.py -q` (244 passed, 1 skipped)
- `uv run pytest --ignore=tests/utils/test_safe_setproctitle.py -q` (402 passed, 2 skipped)
- `uv run pre-commit run --all-files --show-diff-on-failure --color=never`

The complete macOS run has one unrelated failure in `test_setproctitle_crashes_after_fork_with_corefoundation_loaded`: the installed `setproctitle` no longer produces the deliberately expected SIGSEGV. That platform-specific sentinel is skipped by the project’s Linux CI.

## AI assistance disclosure

This change was implemented with OpenAI Codex assistance. The implementation and test output were checked during the assisted session. The PR is intentionally left as a draft until the contribution author completes their personal review of the final diff, as requested in the issue discussion.


nabi-noor:
@boxed — GitHub does not allow an outside-fork contributor to add a formal review request here. Once this draft is marked ready after the contribution author’s final review, would you please review it?

boxed:
Well.. truth be told I'm not really the main author of most of the code anymore heh. I'm more of the guy who got the ball rolling and am enjoying watching how great @Otto-AA and @nicklafleur are :P

nicklafleur:
happy to review when you are ready

